PR #25, opened 2018-10-03
More teleportation tasks

daniel-melia:
Added some extra tasks. 1.5-1.7 to make more tangible for the coder the fact that Bob's qubit ends up in exactly the same state as Alice's message qubit, encouraging them to write a test to show this. 4.1-4.2 are meant to be a bit more challenging and demonstrate that quantum teleportation can occur in arrangements other than a Bell pair shared by two parties.

### DIFF
--- a/Teleportation/ReferenceImplementation.qs
+++ b/Teleportation/ReferenceImplementation.qs
@@ -60,6 +60,31 @@ namespace Quantum.Kata.Teleportation
         }
     }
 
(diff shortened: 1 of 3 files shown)


tcNickolas:
Actually, I take that back: let's merge the PR as is, and worry about coming up with a perfect name for part 4 when we have more protocolrs in it :-)

PR #26, opened 2018-10-03
Included Video lecture for Simon's Problem.

strongSoda:
It might not be comfortable for all learners to go through texts, but a video on the same topic may seem dearer to their eyes.


UnXman:
Arrrr! AI video combined with text makes a good Matrix Great... In to the Shroud!

PR #35, opened 2018-10-20
Joint Measurements kata

tcNickolas:


tcNickolas:
Yes, these are settings which improve the experience for VS Code users.

PR #38, opened 2018-10-22
Extend the description of Simon's algorithm in the tasks file.

lmrazorea:


tcNickolas:
We appreciate you taking the time to write such a detailed description of the Simon's problem. But we don't think that adding very detailed descriptions of the problems the algorithms solve to the katas is the right thing to do. 

Expressive capabilities of Wikipedia, lecture notes or books are vastly superior to those of the comments in Q# files. The Unicode support we have for presenting the tasks cannot really replicate proper LaTeX rendering or images :-) So we would prefer to provide a minimal description for this task and to send the reader to external sources for a more detailed description, as well as the algorithm explanation, consideration of special cases, worked examples etc. - especially since this is a very well studied algorithm, which is covered in practically every course on quantum computing.

PR #39, opened 2018-10-22
Refactor duplicate code in Grover's algorithm kata.

lmrazorea:


GroversAlgorithm/ReferenceImplementation.qs
```diff
@@ -36,19 +36,22 @@ namespace Quantum.Kata.GroversAlgorithm
...
+            FlipOddPositionBits_Reference(queryRegister);
+        }
+        adjoint auto;
+    }
+
+    operation FlipOddPositionBits_Reference (register : Qubit[]) : ()
+    {
+        body
+        {
+            for (i in 0..Length(register) - 1)
```
tcNickolas: If we're really looking to improve this fragment of code, this should be
```suggestion
            for (i in 1..2..Length(register) - 1)
```
This way we don't need the nested `if`, the loop will iterate over odd values of `i` only

GroversAlgorithm/ReferenceImplementation.qs
```diff
@@ -36,19 +36,22 @@ namespace Quantum.Kata.GroversAlgorithm
...
-            {
-                if (n % 2 == 1)
-                {
-                    X(queryRegister[n]);
-                }
-            }
+            FlipOddPositionBits_Reference(queryRegister);
             (Controlled X)(queryRegister, target);
-            for (n in 0..Length(queryRegister)-1)
+            FlipOddPositionBits_Reference(queryRegister);
```
tcNickolas: This should actually be 
```suggestion
            (Adjoint FlipOddPositionBits_Reference)(queryRegister);
```
(and the operation should be declared as having an adjoint), since semantically the second bitflip is duplicate code but uncomputing the effect of the previous bitflip

GroversAlgorithm/ReferenceImplementation.qs
lmrazorea: Using a range expression with a step of 2 gives exception:
System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.

The bug is not as simple as an off-by-one error though.
I've run some tests by adding this loop with an empty range to the code:
```
for (k in 1..2..0) {
	Message("should not reach this statement");
	//X(register[0]);
}
```
The message gets printed but only if the X operation is uncommented.
When commenting out X, the message doesn't get printed. Very strange! This reminds me of C/C++ memory corruption bugs. 
This might actually be a bug in the Quantum Development Kit.
This behavior persists even after removing all additional code and just keeping the Oracle_AlternatingBits task and one test for it.

tcNickolas: Interesting, let me look into it

tcNickolas: This indeed looks like a bug in the way the adjoint of a for loop over a range with a step is generated... We'll have to leave this particular improvement until later version in which this would be fixed.

Thank you for finding this!

PR #43, opened 2018-10-30
Migrate the katas to QDK 0.3 and .NET Core 2.1

tcNickolas:


tcNickolas:
Note that this PR does not take full advantage of the nicer language features introduced in 0.3, like ternary operator or single qubit allocation. There will be fewer of these changes, and they can be done in a separate pass.

PR #45, opened 2018-10-31
Update quick reference to Q# 0.3

tcNickolas:


tcNickolas:
My goal for this quick reference is to be helpful for beginners and workshop attendees, rather than to be a complete syntax guide. It is intentionally incomplete - for example, I didn't include borrowing qubits, partial application, templates, joint measurements and other features. Unwrap operator is definitely not a basic one - it doesn't come up in workshops or katas at this point. So I assume that by the time a user needs this feature they are comfortable enough with the main documentation.

PR #50, opened 2018-11-30
Add CI step to check Unicode symbols used for bra and ket

tcNickolas:
As pointed in #4, there are two symbols that can be used for each of the bra and ket notations. 
The preferred symbols to use are U+27E9 MATHEMATICAL RIGHT ANGLE BRACKET and U+27E8 MATHEMATICAL LEFT ANGLE BRACKET.

This change adds a check to prevent U+232A RIGHT-POINTING ANGLE BRACKET and U+2329 LEFT-POINTING ANGLE BRACKET from creeping back in the code. It also fixes several occurrences of these symbols in the katas.

### DIFF
--- a/scripts/fix-unicode.ps1
+++ b/scripts/fix-unicode.ps1
@@ -1,0 +1,9 @@
+﻿$files = Select-String -Path "**\*.qs" -pattern "[〉〈]" | Select-Object -Unique Path
+
+$files | ForEach-Object {
+    # Get-Content by default reads file in UTF8 encoding
+    $Content = (Get-Content -Encoding UTF8 $_.Path) | ForEach-Object { $_ -Replace "〉", "⟩" } | ForEach-Object { $_ -replace "〈", "⟨" }
+    # Make sure the file is written without BOM
+    $Utf8NoBomEncoding = New-Object System.Text.UTF8Encoding $False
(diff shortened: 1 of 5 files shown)


tcNickolas:
An example of this check failing can be found at https://dev.azure.com/michaylova/Quantum%20Katas%20CI/_build/results?buildId=60.

scripts/fix-unicode.ps1
```diff
@@ -0,0 +1,9 @@
+﻿$files = Select-String -Path "**\*.qs" -pattern "[〉〈]" | Select-Object -Unique Path
+
+$files | ForEach-Object {
+    # Get-Content by default reads file in UTF8 encoding
+    $Content = (Get-Content -Encoding UTF8 $_.Path) | ForEach-Object { $_ -Replace "〉", "⟩" } | ForEach-Object { $_ -replace "〈", "⟨" }
```
cgranade: It may be worth not specifying the encoding explicitly here, just in case someone accidentally introduces a UTF-16 encoded file or something.

tcNickolas: I tried it without encoding as well, and (at least on my machine) by default `Get-Content` does non-UTF8 encoding, so it doesn't recognize the characters to replace. I'm not 100% happy with this solution but since it's not a mandatory tool to use (one can use Visual Studio or VS Code or whatever tool they prefer) I think it's fine.

PR #51, opened 2018-12-14
fix typo in measurements excersice 2.1

krusek:
There was a bit of a calculation error in the description of the solution for 2.1.

|+> = 1/sqrt(2)[1,1] and so Ry(2alpha)*|+> is different.
Also there is a factor of 1/2 missing from P_correct.

As far as I can tell I've done all my calculations correctly.


tcNickolas:
I don't think your formula for E_b is correct. E_a and E_b are obtained by rotating the basis vectors (so that they are orthogonal and we can measure in the basis (E_a, E_b)), not the vectors we need to distinguish, so E_b has to be a rotation of |1⟩, not of |+⟩.

P_correct needs a 1/2 multiplier indeed, good catch!

I reverted the change for E_b and clarified the math a little.

krusek:
Ahh, I see. Not super familiar with von Neumann measurement (very new to quantum computation). It looks like generally I arrived at the same result just through a different order of multiplication. Ie I did (Ry*|+>)*|1> vs (Ry*|1>)*|+>
At least that's what it looks like to me.

tcNickolas:
Yes, I believe this was the case. All part of the learning :-)

PR #58, opened 2019-01-10
Use ASCII for letter variable names

pgarrison:
The script r and x didn't render on my machine, so I'm suggesting they become ASCII.

### DIFF
--- a/DeutschJozsaAlgorithm/Tasks.qs
+++ b/DeutschJozsaAlgorithm/Tasks.qs
@@ -122,7 +122,7 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
     }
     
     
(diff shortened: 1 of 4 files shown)


tcNickolas:
Agreed, if these characters don't render well everywhere it's better to use ASCII instead.

Could you please replace these characters in other katas as well, so that there is just one change fixing this? Looks like there are only few other places which use these, Tasks.qs in `QEC_BitFlipCode` and Tasks.qs and ReferenceImplementation.qs in `SimonsAlgorithm`.

It probably makes sense to add a check for these characters in `scripts/validate-unicode.ps1` (updating the error message accordingly) and a replacement logic in `scripts/fix-unicode.ps1`, so that they don't creep back in the code. (This is optional, I can do this change separately later)

PR #73, opened 2019-02-25
UnitaryPatterns: add tasks from the warmup round

tcNickolas:
These tasks have been featured in Microsoft Q# Coding Contest - Winter 2019 - Warmup (https://codeforces.com/contest/1115)

### DIFF
--- a/UnitaryPatterns/ReferenceImplementation.qs
+++ b/UnitaryPatterns/ReferenceImplementation.qs
@@ -45,4 +45,28 @@ namespace Quantum.Kata.UnitaryPatterns {
         ApplyToEach(H, qs[1 .. Length(qs) - 1]);
         X(Head(qs));
     }
+    
+    
+    // Task 6. Anti-diagonal
+    operation Antidiagonal_Reference (qs : Qubit[]) : Unit {
+        ApplyToEach(X, qs);
+    }
+
+
+    // Task 7. 2x2 chessboard pattern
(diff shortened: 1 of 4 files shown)


UnitaryPatterns/ReferenceImplementation.qs
```diff
@@ -45,4 +45,28 @@ namespace Quantum.Kata.UnitaryPatterns {
...
+    // Task 6. Anti-diagonal
+    operation Antidiagonal_Reference (qs : Qubit[]) : Unit {
+        ApplyToEach(X, qs);
+    }
+
+
+    // Task 7. 2x2 chessboard pattern
+    operation ChessPattern2x2_Reference (qs : Qubit[]) : Unit {
+        H(Head(qs));
+        for (i in 2 .. Length(qs) - 1) {
```
alan-geller: For more than 3 qubits, does this have to skip every other? That is, should the range be 2..2..Length(qs)-1, or what you have? I haven't worked it out, yours might be correct; they're only different for 5 or more qubits.

tcNickolas: No, it only skips the second qubit, after that all qubits get an H. 
The tensor product we're trying to get is H ⊗ ... ⊗ H (a square matrix filled with non-zeroes) ⊗ (I ⊗ H) (a 4x4 block of the shape below), 

        XX..
        XX..
        ..XX
        ..XX

You can see a more detailed solution at https://assets.codeforces.com/rounds/1115/warmup-editorial.pdf

PR #76, opened 2019-02-27
SuperPosition: WState_Arbitrary reference

charlesastaylor:
* Add reference solution without recursion

The quantum computation principle for this solution is exactly the same as the reference already there but as it's iterative and not recursive thought there was a chance it might be worth including.


charlesastaylor:
Thank you (and whoever else is responsible for them) for running your awesome competitions!

PR #82, opened 2019-03-21
[Graph Coloring] Add Kata Implementation

Dan-Tran:
This commit implements a Kata exploring using Grover's search to solve
the constraint satisfaction problem of graph coloring, coloring nodes of
a graph in such a way that no adjacent nodes share the same color.

This project was done as part of UW CSE 490 Q as its final project.
The authors are Dan Tran, Daniel Snitkovskiy, and Frederik Schmitt.


Dan-Tran:
Thank you for the feedback! I really appreciate it. Unfortunately, due to school and a relatively busy schedule, I'll be unable to keep working on it. I hope that my work though is a good stepping stone for the development of the kata! Thank you for the opportunity.

PR #85, opened 2019-03-22
Add CHSH Game kata

marshallsa:
Based on https://github.com/Microsoft/Quantum/tree/master/Samples/src/CHSHGame

The kata also includes an experiment for Bob's qubit rotations (described in the README) that we wrote for our project, but I'm not sure if that should stay in the kata.


tcNickolas:
This looks good overall, thank you! Here are several high-level comments/ideas for improvement:

- I think you're right doubting about the experiment files, `Experiments.qs` and `Driver.cs` don't really belong to the kata, so it's better to remove them.

- It would be nice to start the kata with a small section on classical strategy - two operations `AliceClassical` and `BobClassical` which each take a bit and return a bit, covered by one test to run them against each other. Another small (and purely classical) task would be the check - given bits A, B, X and Y, check whether they satisfy the win condition of the game - to reinforce what is it that we're trying to do here.

- Task 5 "Play the CHSH game" is not really a good representation of the game - it gives the solution full information about both input bits, and the solution can get 100% win rate easily (it's just the test that prevents this from happening). On the other hand, it's hard to represent it in a different way, since you've already covered the individual pieces of strategies in previous tasks...
You can try to have task 5 accept arguments of two operations, each of them with a signature `(Qubit => Bool)`, instead of bits A and B. Then you can choose bits A and B given to the strategies in the test harness, and pass to task 5 operations `MeasureAliceQubit(A, _)` and  `MeasureBobQubit(B, _)` - this way you can deny task 5 the knowledge of the actual bits while still asking it to do the qubit allocation, the entangled pair preparation and the use of strategies, i.e., a big part of the interaction protocol.

Please let me know if you want to do these changes; if not, I'll probably have some time for that after I've done a first pass on the rest of the PRs, maybe next week.

marshallsa:
Thanks for the feedback! Those sound like good suggestions. I agree that task 5 can be done better, in particular. I'll try to find some time to make the changes (hopefully by this weekend).

marshallsa:
@tcNickolas Changes made. :)

Should the comment from the [CHSH game sample](https://github.com/Microsoft/Quantum/blob/master/Samples/src/CHSHGame/Game.qs) be included in this kata (in the Tasks.qs header or the README) as an introduction for the users?

tcNickolas:
Wow, that was fast! :-) I'll take a look on Monday.

I would include a very brief introduction (literally a couple of words about what nonlocal games show and how CHSH game is one of the most famous examples of them) to the README.md, but not copy the whole comment. I prefer including sources, like https://cs.uwaterloo.ca/~watrous/CPSC519/LectureNotes/20.pdf, whenever possible instead of full explanations.

marshallsa:
> * tests for tasks 2.2 and 2.4 - I think it's better not to restrict the state of the qubit after the measurement, since there are different measurements that can produce the same answer but leave the qubit in different states.

I did this to try to make sure they weren't measuring in the wrong basis and just getting lucky with the results, but I think I see how it's too strict now (if they use H then measure in the Z basis), especially since measurements in the wrong basis should still be caught by the test for task 2.5.

The changes look good. :)

PR #86, opened 2019-03-22
Add GHZ Game kata

marshallsa:
The naming conventions and task descriptions might need to be changed to match the other katas.


marshallsa:
Thank you for the detailed comments! I'm not sure if we had more tests planned or if we forgot to remove the comment after adding the tests. :) @justinmsievers do you remember?

I can make the changes at some point (probably after CHSH game) unless Justin wants to.

justinmsievers:
> Thank you for the detailed comments! I'm not sure if we had more tests planned or if we forgot to remove the comment after adding the tests. :) @justinmsievers do you remember?
> 
> I can make the changes at some point (probably after CHSH game) unless Justin wants to.

@marshallsa  Yeah, we implemented all planned tests. This was a forgotten note.

marshallsa:
@tcNickolas I've made most of the changes. I have a couple questions:

* `PlayClassicalGHZ` is given all of the inputs right now, so it would be possible for it to win every time (other than that the test would keep it at 50% or 75% depending on the strategy). But it doesn't really make sense to give it the same treatment as `PlayQuantumGHZ` and `PlayQuantumCHSH` since there aren't any qubits, so it would just be calling a zero-argument function 3 times. Should we let the task see all the inputs or remove the task completely?
* Task 2.2 is missing a test right now (it's tested as part of task 2.3). The task description is vague about what the quantum strategy is. Should we just say that they should apply H if the bit is 1, like the CHSH kata did? Then we can test that `QuantumStrategy(true, _)` is equivalent to H and `QuantumStrategy(false, _)` is equivalent to identity (kind of trivial, but the quantum strategy for GHZ is that simple).

Thanks!

tcNickolas:
* For `PlayClassicalGHZ`, I think it's fine to give all input bits if we're providing the strategies the solution needs to use - the goal of this task is not to win the game but to run the interaction between players (referee side implementation).
* For task 2.2, I think it makes sense to follow the same approach as in CHSH kata, describing the strategy exactly.

I went ahead and added the tests for these two tasks. I've also done some cleanup, including borrowing the `fail "Task not implemented yet";` trick from your CHSH kata - it makes a lot of sense for simple strategies and for tasks which rely on other tasks being implemented.

I'm happy with the state of this kata - please take a look, and if everything looks good to you too I'll merge it.

Thank you!

marshallsa:
For task 1.4, they're given an array and need to return another array, but for task 1.1 and task 2.3 they're given a tuple and need to return another tuple (except task 1.1 which only returns a Bool). I'm not sure which is better but I think it would make sense to pick either arrays or tuples for all the tasks for consistency. I like tuples since you can name each element but arrays are more concise (and maybe more consistent with qubit registers like in task 2.1).

The rest of the changes look good!

marshallsa:
Looks good!

PR #87, opened 2019-03-22
Add Mermin-Peres magic square game kata

marshallsa:
Like the GHZ game kata there might need to be some changes to the operation naming conventions and task descriptions (sorry).

There's also an interactive version that you can run to play as the adversary who picks the row and column for Alice and Bob (described in the README).


marshallsa:
> 
> 
> [CLA assistant check](https://cla.opensource.microsoft.com/Microsoft/QuantumKatas?pullRequest=87)
> All CLA requirements met.

Do the co-authors need to sign the CLA too?

tcNickolas:
@marshallsa No, the CLA bot is happy with just one person signing CLA. It says that you have the right to grant Microsoft the right to use your code, so I guess in case of the contributors it implies that you've discussed this with the contributors offline and decided that it's ok :-)

Sorry for the delay reviewing the PRs, a lot of them came in at once so it will take us some time to go through them all.

marshallsa:
Thanks again for the extensive feedback! Sorry for the messy katas, the last couple weeks of the quarter were super busy and we were just trying to get everything done on time. :) I'll try to find time to make these changes, probably also after I do CHSH game (since that that looks like it's the closest to ready).

tcNickolas:
No worries, I know the feeling of squeezing a lot of work in a little time :-) I agree, I would start with CHSH, followed by GHZ, followed by this one.

tcNickolas:
I was thinking more about how to keep the interactive game in the kata - it is cute indeed, and it is nice to play as the opponent once in a while :-) This is similar to the last tasks in the other two katas - you have to pick the input and the strategy, and use them to get the results of the game). 

We can make it into a new task (the last one for the kata), convert console prompts into comments in the task, and don't cover it with tests, like [task 3.2 in GroversAlgorithm kata](https://github.com/Microsoft/QuantumKatas/blob/master/GroversAlgorithm/Tasks.qs#L212) or [task 2.1 in SolveSATWithGrover kata](https://github.com/Microsoft/QuantumKatas/blob/master/SolveSATWithGrover/Tasks.qs#L205). This way we keep the experimentation angle (and visualization), but also keep it open-ended so that there is no single "right" way to do it.

What do you think?

marshallsa:
I think that's a good idea, so the interactive prompt for a strategy would translate to picking which strategy operation to call, and the prompt for row and column would be passing a row/column as an argument to the operations? That would probably work better than the console program since it's easier to run multiple trials, randomize the inputs, etc. like we do in the tests.

Would the visualization just be giving them our grid drawing function to call in the test if they want to or did you have something else in mind?

tcNickolas:
Yes, exactly. For visualization I'd just point them to the function they can use, like [this task](https://github.com/Microsoft/QuantumKatas/blob/master/DeutschJozsaAlgorithm/Tasks.qs#L277) does. Or maybe give them a skeleton of what it would look like to call classical strategy with parameters 0, 0 and visualize the result, and some suggestions on what they can try to tweak to make the players lose.

marshallsa:
A couple of changes so far:

* I removed the suboptimal classical strategy tasks since they didn't seem very well-defined (the solution is to always output the same signs for every row/column, but we only tell them to achieve a win rate of 66% in the task). I kept the optimal classical strategy tasks since those are well-documented in sources about the game.
* I tried adding a test for task 2.2 (magic square observables) but the test ended up being pretty complicated and I'm not sure how effective it actually is. I think there are multiple configurations of operators that work so we probably shouldn't require that they use a particular one. I tried checking the [properties of the square](https://en.wikipedia.org/wiki/Quantum_pseudo-telepathy#The_Mermin-Peres_magic_square_game) (all observables in each row/column commute, product of observables in a row is I, product of observables in a column is -I), but I don't think `AssertOperationsEqualReferenced` can tell the difference between I and -I. So right now the trivial case where every cell is I ⊗ I will pass the test. Also I'm not sure but I think even if all the properties are satisfied it doesn't necessarily mean the strategy will win the game, so maybe we might have to leave this to be tested as part of the win rate test?

marshallsa:
@tcNickolas I've made most of the changes now. :) I added a task for experimenting with the strategies and a `DrawMagicSquare` function to show what the magic square looks like. I also added a test for task 2.3, but I'm not sure if there's a better way to do it (maybe something like `AssertOperationsEqualReferenced` that also checks the return value?).

Things I haven't gotten to are adding an intro about the game rules to Tasks.qs, updating the README, and fixing/maybe removing the test for task 2.2.

tcNickolas:
Sorry for the delay reviewing - this one is more complicated compared to the other two games, so I need to spend more time figuring out the math behind the quantum section of the tasks. So far I updated the project to 0.6 and cleaned up the tasks up to 2.1.

I think `AssertOperationsEqualReferenced` on its own doesn't recognize global phase, and there are definitely different possible solutions to task 2.2 (Wikipedia and assignment solutions give different squares of operators - at the very least you have the freedom to reorder rows and columns), so it might be tricky to test indeed. We might want to drop 2.2 and replace it with a hint in task 2.4. I'll keep thinking about it.

I'll also think some more about 2.3 - a similar thing can be done using joint measurements, though we'd need to use Pauli matrices as an argument instead of an operation. I need to check how the signs behave in both implementations and whether they make a difference.

marshallsa:
Thanks for cleaning up the code! The changes look good.

Yes, it looks like the easiest thing to do is combine tasks 2.2 and 2.4. That would make task 2.4 pretty big since it already has Alice and Bob's strategies combined though. Maybe there is some way to hook up each observable as part of a larger circuit that we can use to see the phase? The phase definitely has some observable effect in the game.

In the Measurement kata, it looks like it checks if a measurement is correct by running 100 trials. We could just do that for task 2.3. :) Right now it only has one attempt for each of the 9 operators.

tcNickolas:
I'm still working on part 2 - turns out I needed to learn some things to properly understand what's happening and to tweak it :-) I'll need more time to clean up the changes, but here is the outline:
* You can represent an observable as a pair of Pauli values plus sign by which they are multiplied. This means we can switch task 2.2 to returning an observable in this format, and spin off a new task - applying a unitary which corresponds to the given observable. (It took me some effort to figure out how to do it formally, and where does the minus sign come in, so I think it's a good task.)
* You can check that the unitary is applied with correct global phase (including the sign) if you apply `AssertOperationsEqualReferenced` to a controlled version of this unitary (controlled version makes global phase into a relative one, which can be verified), so it is possible to check the return in task 2.2 more exactly. And I think that if the observables satisfy the conditions you listed, they will be a solution to the game, so checking the conditions should be fine.
* The task of measuring an observable becomes two tasks as well - one task to measure the given unitary like you did, and another one to measure an observable described as a pair of Paulis (you can use joint measurements to solve the second task, and again you have to be careful with the sign, so hopefully this task will teach a bit more about the signs of observables).

I'll try to finish this on Monday or Tuesday - I really like how it looks so far!

marshallsa:
Sounds great, looking forward to seeing the rest of the changes!

tcNickolas:
Here are the changes for the second part of the kata - looks good to me. Let me know what you think!

marshallsa:
Looks good! I was wondering what the `ApplyMagicObservables` and `MeasureOperator` tasks are for? It looks like they can implement their strategy using only `MeasureObservable` instead of those two. Is converting the observable to an operator and measuring the operator meant to be an alternative way to do the quantum strategy?

tcNickolas:
Yes, if one doesn't have `Measure` operator for doing joint measurements, that's how they would be measuring those observables - I spent some time figuring this out so I think it's a good thing to learn. Knowing several ways to accomplish a task is always educational.

marshallsa:
Should we add a hint to task 2.6 saying that instead of `MeasureObservable`, they can also use `ApplyMagicObservables` together with `MeasureOperator`? Also we could show how both are done in the reference solution? (Maybe Alice could use `MeasureObservable` and Bob could do it the other way.)

tcNickolas:
Good point, there's no need for two strategies to do the same thing, they can show different approaches. 
Do you want to do this change? If not, I'll get to it tomorrow morning.

marshallsa:
I changed Bob's strategy to use `MeasureOperator` and added some comments/hints about it. :)

tcNickolas:
Looks perfect now :-) I merged the PR - thank you again!

PR #96, opened 2019-04-19
Add infrastructure for Katas as Jupyter Notebooks

tcNickolas:
* Add Microsoft.Quantum.Katas project that contains the infrastructure necessary for representing Katas in Jupyter Notebook format. 
* Add Docker image to run Notebooks on Binder.
* Convert BasicGates and part of DeutschJozsaAlgorithm katas to Notebook format.
* Update QDK to 0.5.1904.1302 version.

Co-authored by: Andres Paz <anpaz@microsoft.com>


Microsoft.Quantum.Katas/CounterSimulator.cs
```diff
@@ -0,0 +1,194 @@
...
+    ///     and the maximum number of qubits allocated at any point during program execution.
+    /// </summary>
+    public class CounterSimulator : QuantumSimulator
+    {
+        private Dictionary<ICallable, int> _operationsCount = new Dictionary<ICallable, int>();
+        private long _qubitsAllocated = 0;
+        private long _maxQubitsAllocated = 0;
+
+        #region Counting operations
+        public CounterSimulator(
```
cgranade: Would you be willing to add a documentation comment here explaining the parameters to this constructor?

Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            catch (Exception e)
+            {
+                channel.Stderr(e.Message);
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Executes the given kata using the provided `userAnswer` as the actual answer.
```
cgranade: C# XML comments don't support the `` ` `` syntax, so far as I'm aware. I think you need `<c>` and </c>` instead:
```suggestion
        /// Executes the given kata using the provided <c>userAnswer</c> as the actual answer.
```

Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            catch (AggregateException agg)
+            {
+                foreach (var e in agg.InnerExceptions) { channel.Stderr(e.Message); }
+                channel.Stderr($"Try again!");
+                return false;
+            }
+            catch (Exception e)
+            {
+                channel.Stderr(e.Message);
```
cgranade: It may be worth taking an `ILogger` in the constructor and logging exceptions here so that we can get better debugging info (including the full stack trace) if something goes wrong.

Microsoft.Quantum.Katas/Utils.qs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//////////////////////////////////////////////////////////////////////
+// This file contains some Q# intrinsic operations
+// needed for the validation of some Katas' answers.
+//////////////////////////////////////////////////////////////////////
+
+namespace Quantum.Kata.Utils {
```
cgranade: If I understand correctly, this namespace isn't part of the katas themselves (that is, doesn't include code written by users during a kata)? If so, I think that the consensus has been that it needs to live under the `Microsoft.Quantum.` prefix.
```suggestion
namespace Microsoft.Quantum.Kata.Utils {
```

Microsoft.Quantum.Katas/Utils.qs
```diff
@@ -0,0 +1,28 @@
...
+    
+    /// # Summary
+    /// Returns how many times a given oracle is executed.
+    function GetOracleCallsCount<'T> (oracle : 'T) : Int { body intrinsic; }
```
cgranade: I think this needs to be an operation, not a function, since its output can depend on global state (that is, it's not deterministic).
```suggestion
     /// # Input
     /// ## oracle
     /// The operation whose call count is being requested.
    operation GetOracleCallsCount<'T> (oracle : 'T) : Int { body intrinsic; }
```

Microsoft.Quantum.Katas/Utils.qs
```diff
@@ -0,0 +1,28 @@
...
+    
+    /// # Summary
+    /// Resets the variable that tracks how many times an oracle
+    /// is executed back to 0.
+    function ResetOracleCallsCount () : Unit { body intrinsic; }
```
cgranade: Similary, these all need to be operations as well, as they are not deterministic.

tcNickolas:
I've addressed most of the comments, the ones which are left look like they require a NuGet package update - @anpaz-msft, could you please take a look at them?

There is one more piece of work left, updating READMEs - writing new ones for Common and M.Q.Katas projects and restructuring existing ones - but I'd rather do it separately after this one is merged.

Microsoft.Quantum.Katas/Utils.qs
anpaz-msft: This would still be a breaking change. I'd rather not do it.

Microsoft.Quantum.Katas/KataMagic.cs
anpaz-msft: I've added the ILogger and use it for unexpected exceptions. Most errors, though, are expected, so not really creating logs for those.
While doing so, clean up the code to remove Exception when user errors are expected.

Issue #92, opened 2019-04-15
SolveSATWithGrover - Task 2.2 - infinite loop


hishikitetsuya:
```
    // Task 2.2. Universal implementation of Grover's algorithm
    operation GroversAlgorithm_Reference (N : Int, oracle : ((Qubit[], Qubit) => Unit : Adjoint)) : Bool[] {
        mutable answer = new Bool[N];
        using ((register, output) = (Qubit[N], Qubit())) {
            mutable correct = false;
            mutable iter = 1;
            repeat {
                Message($"Trying search with {iter} iterations");
                GroversAlgorithm_Loop(register, oracle, iter);
                let res = MultiM(register);
                // to check whether the result is correct, apply the oracle to the register plus ancilla after measurement
                oracle(register, output);
                if (MResetZ(output) == One) {
                    set correct = true;
                    set answer = BoolArrFromResultArr(res);
                }
                ResetAll(register);
            } until (correct) 
            fixup {
                set iter = iter * 2;
            }
```

It looks like the reference code never stops when we face unlucky case. The Grover's algorithm is not deterministic algorithm and the lower bound of possibility where we can obtain correct answer for unknown number of solutions is 0.5 when we increase k = 1,2,...,n for 2^k iterations. 

Do we need to add upper bound for the loop? Or, do we need add description on the task such as "the given problem can be solved 100% with 2^k grover iterations" ?

If my understanding is not correct, please close the issue.

tcNickolas:
My intention for this task was to show that you can use Grover search to solve the problem even if you don't know how many solutions it has (and thus can not estimate the "correct" number of iterations to use). The exact sequence of the numbers of iterations can vary - you can try 1, 2, 3 etc. iterations, or you can try powers of 2, or you can try the number of iterations you'd do if there was 1 solution, 2 solutions, 3 solutions, etc. Even if you don't hit the "correct" number of iterations in your sequence, you will eventually loop around to a number that has high enough probability of success.

Does that make sense? We can add a note to this effect to the task description if we come up with a good wording - I don't want to promise that the problem can be solved with 2^k iterations, since even if you apply the "correct" number of iterations and maximize the probability of success, you still can get unlucky and measure a state that is not a solution.

hishikitetsuya:
Thanks for rephrasing and that's my point. I was wondering if we could add some sentences as a note for this task.

tcNickolas:
[...]

PR #98, opened 2019-04-22
SolveSATWithGrover: Clarify the solution for task 2.2

tcNickolas:
This change addresses issue #92:
* describe the intent behind the task and the logic of the reference solution, 
* add an extra exit condition to prevent infinite loop when trying increasing numbers of iterations.


SolveSATWithGrover/Tasks.qs
```diff
@@ -222,6 +222,11 @@ namespace Quantum.Kata.GroversAlgorithm {
     // Output:
     //      An array of N boolean values which satisfy the expression implemented by the oracle
     //      (i.e., any basis state marked by the oracle).
+    // 
+    // Note that the similar task in GroversAlgorithm kata required you to implement Grover's algorithm
```
alan-geller:
```suggestion
    // Note that the similar task in the GroversAlgorithm kata required you to implement Grover's algorithm
```

PR #115, opened 2019-05-22
Updated to use Jupyter Notebooks

jackhyder:


tcNickolas:
I did a bit of cleanup on the notebook, here is a quick summary of the most important points:

* Spaces instead of tabs - if you edit .ipynb file manually (say, in Visual Studio), most JSON editors will replace spaces with tabs, but Jupyter Notebook serialization defaults to spaces. Fortunately, this is easy to fix - the first time you edit the notebook in the web-interface and save it, the tabs will be replaced by spaces again. It is generally better to do editing in the web-interface. (My commit will look like I edited every line of the ipynb file, but this is not true :-))

* The welcome instructions slightly differ between the project and the notebook formats - the project instructions mention running tests, while the notebook instructions explain how to execute a cell. It is easiest to copy the last part of the instructions from another kata notebook.

* I like to include the reference materials from README.md to the introduction section of the notebook, since if somebody works through the kata using Notebooks interface, they might never see the README.md.

* `%workspace reload` cell is necessary for the katas which use the custom simulator, like DeutschJozsaAlgorithm, but this one uses the standard simulator so it will be fine without this cell.

* Markdown doesn't recongnize numbered lists with 1) 2) numbering, you have to use 1. 2. to enumerate inputs.

* In task 2.1, using `EqualityFactI` requires opening the namespace in that cell (`open Microsoft.Quantum.Diagnostics`). I removed this assertion altogether - it seems an overkill for such a lightweight task.

* In task 2.3, I like to move hints out from the Q# cell to the Markdown cell above it, so that they can be hidden until clicked on - you can see how we use `<details>` HTML tag for that.

* The QDK reference in .csproj file duplicates an earlier one, no changes should be required for that file.

* I added link to the notebook version of the kata to its readme and to the index.ipynb file, like we did for the rest of the kata notebooks.

It looks like there is a problem with the notebook not being able to recognize previously implemented tasks - for example, if we keep `AliceClassical` and `BobClassical` in the same cell, only the first one is recognized, but if we move them to different cell and keep the test on the second one, only the second one is recognized. I couldn't find a way to make sure both operations are recognized at the same time - and we'll have the same issue with `PlayQuantumCHSH` which implies that both `AliceQuantum` and `BobQuantum` are implemented :-( I'm following up on this, will let you know when I figure out the right way to do that.

PR #116, opened 2019-05-23
Let the kata notebooks override all attempted tasks

tcNickolas:
In https://github.com/microsoft/QuantumKatas/pull/115 we observed that a kata notebook only overrides the task that is currently tested, and uses stubs from the project for the rest of the tasks. This prevents us from implementing tasks which require two related operations be implemented at once, or ones which rely on previously completed tasks.

This change adds tracking of previously attempted tasks and overriding them together with the currently attempted task.

### DIFF
--- a/utilities/Microsoft.Quantum.Katas/KataMagic.cs
+++ b/utilities/Microsoft.Quantum.Katas/KataMagic.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Extensions.Logging;
 using Microsoft.Jupyter.Core;
@@ -27,10 +28,11 @@ namespace Microsoft.Quantum.Katas
             this.Resolver = resolver;
             this.Snippets = snippets;
             this.Logger = logger;
+            this.AllAnswers = new Dictionary<OperationInfo, OperationInfo>();
         }
 
         /// <summary>
-        /// The Resolver let's us find compiled Q# operations from the workspace
+        /// The Resolver lets us find compiled Q# operations from the workspace
         /// </summary>
         protected IOperationResolver Resolver { get; }
 
@@ -40,6 +42,8 @@ namespace Microsoft.Quantum.Katas
         protected ISnippets Snippets { get; }
 
         protected ILogger<KataMagic> Logger { get; }
+
+        protected Dictionary<OperationInfo, OperationInfo> AllAnswers { get; }
 
         /// <summary>
         /// What this Magic does when triggered. It will:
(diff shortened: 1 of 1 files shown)


utilities/Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -178,8 +186,17 @@ public virtual bool Simulate(OperationInfo test, OperationInfo userAnswer, IChan
         /// It does this by finding another operation with the same name as the `userAnswer` but in the 
         /// test's namespace
         /// </summary>
-        public virtual OperationInfo FindSkeletonAnswer(OperationInfo test, OperationInfo userAnswer) =>
-            Resolver.Resolve($"{test.Header.QualifiedName.Namespace.Value}.{userAnswer.FullName}");
+        public virtual OperationInfo FindSkeletonAnswer(OperationInfo test, OperationInfo userAnswer)
+        {
+            var skeletonAnswer = Resolver.Resolve($"{test.Header.QualifiedName.Namespace.Value}.{userAnswer.FullName}");
+            Console.Out.WriteLine($"Resolved {userAnswer.FullName} to {skeletonAnswer}");
```
anpaz-msft: Let's use `Logger.Debug` or `Logger.Info` instead.

utilities/Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -178,8 +186,17 @@ public virtual bool Simulate(OperationInfo test, OperationInfo userAnswer, IChan
...
+            if (skeletonAnswer != null)
+            {
+                // Remember the last user answer for this task
+                AllAnswers[skeletonAnswer] = userAnswer;
```
anpaz-msft: Have you confirmed that `OperationInfo` works well as key for dictionary? I think it does, but it would be nice to make sure that if you run the same cell multiple times indeed the latest userAnswer is the one used.

utilities/Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -137,7 +141,11 @@ public virtual bool Simulate(OperationInfo test, OperationInfo userAnswer, IChan
                 var qsim = CreateSimulator();
 
                 qsim.DisableLogToConsole();
-                qsim.Register(skeletonAnswer.RoslynType, userAnswer.RoslynType, typeof(ICallable));
+                // Register all solutions to previously executed tasks (including the current one)
+                foreach (KeyValuePair<OperationInfo, OperationInfo> answer in AllAnswers) {
+                    Console.Out.WriteLine($"Registering {answer.Key.FullName}");
```
anpaz-msft: Same here, please us `Logger.Debug` for this one.

utilities/Microsoft.Quantum.Katas/KataMagic.cs
tcNickolas: Yes, I verified it using the pair of Alice/Bob strategies from #115 - the latest solution from each cell is used.

PR #122, opened 2019-05-31
Updating to 0.7.1905.3109

anpaz-msft:


BasicGates/BasicGates.ipynb
```diff
@@ -36,7 +36,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%package Microsoft.Quantum.Katas::0.6.1905.301"
+    "%package Microsoft.Quantum.Katas::0.7.1905.3109"
```
tcNickolas: Could you please update the package versions mentioned in the "How to install the right IQ# version" section in all the Notebooks? It makes sense to keep them in sync with the rest of the versions, otherwise it will be confusing.

anpaz-msft: Can I change the description to be a generic version number? I have a script to update version numbers, but that is a bit random text so I have to manually update those and I don't want to do it on every release.

anpaz-msft: Replaced the text with `0.1.2.3` so we don't have to update it on every release... hope that is enough.

tcNickolas: It should be possible to modify the script to replace all occurrences of the old version with the new one in Notebooks. It can be dangerous in general with shorter versions, if the version is 0.1 it will get replaced in floating-point constants, but with the versioning scheme we're using I find it hard to come up with a counter-example in Notebooks...

I'm fine with this for now, but in general it is more convenient for the user to have the correct version in installation instructions - I'm pretty sure people will copy-paste command with 0.1.2.3 and be surprised when it doesn't work.

PR #130, opened 2019-06-19
Beheim/superdense coding

bettinaheim:
I did some minor updates to the superdense coding kata related to a demo. 
In particular, I switched to using a user defined type Message to represent the two bits of classical information to transmit.


SuperdenseCoding/Tasks.qs
```diff
@@ -3,11 +3,16 @@
...
     open Microsoft.Quantum.Intrinsic;
-    open Microsoft.Quantum.Canon;
-    
+    open Microsoft.Quantum.Measurement;
     
+    /// # Summary
+    /// Data type that represents the message that is transmitted 
+    /// as part of the superdense coding protocol. 
+    /// It includes two classical bits (Bit1 and Bit2) of type Bool.   
+    newtype Message = (Bit1 : Bool, Bit2 : Bool);
```
cgranade: Minor thought, but do we want to use the Alice and Bob naming scheme here as well?

SuperdenseCoding/Tests.qs
```diff
@@ -14,57 +14,50 @@ namespace Quantum.Kata.SuperdenseCoding {
     
     
     // ------------------------------------------------------
-    operation AssertEqualOnZeroState (N : Int, taskImpl : (Qubit[] => Unit), refImpl : (Qubit[] => Unit is Adj)) : Unit {
-        using (qs = Qubit[N]) {
+        
+    operation T1_CreateEntangledPair_Test () : Unit {
+        using ((q1,q2) = (Qubit(), Qubit())) {
```
cgranade:
```suggestion
        using ((q1, q2) = (Qubit(), Qubit())) {
```

SuperdenseCoding/Tests.qs
```diff
@@ -14,57 +14,50 @@ namespace Quantum.Kata.SuperdenseCoding {
...
+            CreateEntangledPair(q1, q2);
+
             // apply adjoint reference operation and check that the result is |0^N⟩
-            Adjoint refImpl(qs);
-            
+            Adjoint CreateEntangledPair_Reference(q1, q2);
+
             // assert that all qubits end up in |0⟩ state
-            AssertAllZero(qs);
+            AssertAllZero([q1,q2]);
```
cgranade:
```suggestion
            AssertAllZero([q1, q2]);
```

SuperdenseCoding/Tasks.qs
tcNickolas: I believe the compiler doesn't mind, but for me type `Message` just reads as logging function `Message` and is confusing... Could we call it `ProtocolMessage` instead?

SuperdenseCoding/Tasks.qs
```diff
@@ -40,22 +45,19 @@ namespace Quantum.Kata.SuperdenseCoding {
     // Task 1. Entangled pair
     // Input: An array of two qubits in the |00⟩ state.
```
tcNickolas: Please update input description to match operation signature.

SuperdenseCoding/Tasks.qs
```diff
@@ -68,21 +70,16 @@ namespace Quantum.Kata.SuperdenseCoding {
     }
     
     
-    // Task 3. Decode the message (Bob's task)
-    // Decode the message using the qubit received from Alice.
+    // Task 3. Decode the message and reset the qubits (Bob's task)
+    // Decode the message using the qubit received from Alice and reset them to a |00⟩ state.
     // Inputs:
     //      1) Bob's part of the entangled pair qBob.
```
tcNickolas: Please update input description (order of inputs) to match operation signature.

SuperdenseCoding/Tasks.qs
```diff
@@ -68,21 +70,16 @@ namespace Quantum.Kata.SuperdenseCoding {
     }
     
     
-    // Task 3. Decode the message (Bob's task)
-    // Decode the message using the qubit received from Alice.
+    // Task 3. Decode the message and reset the qubits (Bob's task)
+    // Decode the message using the qubit received from Alice and reset them to a |00⟩ state.
```
tcNickolas:
```suggestion
    // Decode the message using the qubit received from Alice and reset both qubits to a |00⟩ state.
```

SuperdenseCoding/Tasks.qs
```diff
@@ -40,22 +45,19 @@ namespace Quantum.Kata.SuperdenseCoding {
...
     }
     
     
     // Task 2. Send the message (Alice's task)
-    // Encode the message (classical bits) in the state of Alice's qubit.
+    // Encode the message (two classical bits) in the state of Alice's qubit.
     // Inputs:
     //      1) Alice's part of the entangled pair of qubits qAlice.
-    //      2) two classical bits, stored in an array.
+    //      2) two classical bits, stored as Message.
```
tcNickolas: This is the first time people see a type, so I'd suggest giving a quick explanation of how they can access the bits (`message::Bit1`).

SuperdenseCoding/Tasks.qs
```diff
@@ -68,21 +70,16 @@ namespace Quantum.Kata.SuperdenseCoding {
...
-        // Declare a Bool array in which the result will be stored;
-        // the variable has to be mutable to allow updating it.
-        mutable decoded_bits = new Bool[2];
-        
-        // ...
+    // The state of the qubits in the end of the operation should be |00⟩.
+    operation DecodeMessageFromQubits (qAlice : Qubit, qBob : Qubit) : Message {
 
-        return decoded_bits;
+        fail ("not implemented");
```
tcNickolas:
```suggestion
        fail ("Task 3 not implemented");
```

SuperdenseCoding/Tasks.qs
```diff
@@ -93,13 +90,9 @@ namespace Quantum.Kata.SuperdenseCoding {
...
-    operation SuperdenseCodingProtocol (message : Bool[]) : Bool[] {
-        
-        mutable decoded_bits = new Bool[2];
-        
-        // ...
-
-        return decoded_bits;
+    operation SuperdenseCodingProtocol (message : Message) : Message {
+                
+        fail ("not implemented");
```
tcNickolas:
```suggestion
        fail ("Task 4 not implemented");
```

SuperdenseCoding/Tasks.qs
bettinaheim: For the message? That to me seems more confusing rather than helpful, since it sounds like it implies that one bit belongs to Alice, and one to Bob.

bettinaheim: Sure.

SuperdenseCoding/Tasks.qs
bettinaheim: I added the first if-statement (without body) to the implementation stump below and commented that it accesses Bit1 of message. Take a look and let me know if that's what you had in mind, or if you have a better idea on how to explain named item access.

SuperdenseCoding/Tasks.qs
bettinaheim: I updated the description. Take a look if you are happy with it.

SuperdenseCoding/Tasks.qs
```diff
@@ -10,7 +10,7 @@ namespace Quantum.Kata.SuperdenseCoding {
     /// Data type that represents the message that is transmitted 
     /// as part of the superdense coding protocol. 
     /// It includes two classical bits (Bit1 and Bit2) of type Bool.   
-    newtype Message = (Bit1 : Bool, Bit2 : Bool);    
+    newtype ProtocollMessage = (Bit1 : Bool, Bit2 : Bool);
```
tcNickolas: Typo: should be `ProtocolMessage` (with one `l`)

SuperdenseCoding/Tasks.qs
```diff
@@ -38,24 +43,22 @@ namespace Quantum.Kata.SuperdenseCoding {
...
     
     // Task 1. Entangled pair
-    // Input: An array of two qubits in the |00⟩ state.
-    // Goal:  Create a Bell state |Φ⁺⟩ = (|00⟩ + |11⟩) / sqrt(2) on these qubits.
-    operation CreateEntangledPair (qs : Qubit[]) : Unit {
-        // The following lines enforce the constraints on the input that you are given.
-        // You don't need to modify them. Feel free to remove them, this won't cause your code to fail.
-        EqualityFactI(Length(qs), 2, "The array should have exactly 2 qubits.");
+    // Input: A tuple of two qubits.
+    // Goal:  If both qubits are initially in a state |00>,
```
tcNickolas: This task description is incorrect/incomplete; we don't ask people to figure out whether the qubits are in state 00 and modify the state only if they are, we ask them to prepare a certain state and guarantee that the qubits start in 00.

If you want to modify this task to ask for a unitary transformation H-CNOT, you need to properly describe it, and to modify the test `T1_CreateEntangledPair_Test` to use `AssertOperationsEqualReferenced` instead of `AssertEqualOnZeroState` to make sure that it checks exactly the same thing as the task asks to do.

If you want to keep this task as is (which, to be fair, is what I would do - we don't *force* people to use adjoint of this task to perform later stages of the teleportation protocol, and formulating the task as state preparation doesn't prevent *us* from doing it our way and using it later), keep the goal as it was and modify input to "Two qubits, each in the |0⟩ state."

(Also, note the use of fancy Unicode |0⟩ instead of plain ASCII |0> to denote states :-) )

SuperdenseCoding/Tasks.qs
```diff
@@ -38,24 +43,22 @@ namespace Quantum.Kata.SuperdenseCoding {
     // Finally, we compose those steps into the complete superdense coding protocol.
     
     // Task 1. Entangled pair
-    // Input: An array of two qubits in the |00⟩ state.
-    // Goal:  Create a Bell state |Φ⁺⟩ = (|00⟩ + |11⟩) / sqrt(2) on these qubits.
-    operation CreateEntangledPair (qs : Qubit[]) : Unit {
-        // The following lines enforce the constraints on the input that you are given.
-        // You don't need to modify them. Feel free to remove them, this won't cause your code to fail.
-        EqualityFactI(Length(qs), 2, "The array should have exactly 2 qubits.");
+    // Input: A tuple of two qubits.
```
tcNickolas:
```suggestion
    // Input: Two qubits.
```

SuperdenseCoding/Tasks.qs
```diff
@@ -64,25 +67,23 @@ namespace Quantum.Kata.SuperdenseCoding {
         // [1; 0]:    |Φ⁻⟩ = (|00⟩ - |11⟩) / sqrt(2)
         // [1; 1]:    |Ψ⁻⟩ = (|01⟩ - |10⟩) / sqrt(2)
 
+		if (message::Bit1) { // accesses the item 'Bit1' of 'message'
```
tcNickolas: Please use spaces instead of tabs.

SuperdenseCoding/Tasks.qs
bettinaheim: Fair point. As suggested, I kept the original task goal and modified the input description to "Two qubits, each in the |0⟩ state.".

bettinaheim:
> I fixed the Notebook to refer to the data type defined in the project instead of defining its own data type (they have different namespaces), and added a few comments to help solvers figure out the syntax of dealing with data types.
> Other than that, looks good to me.

Awesome, thanks!

PR #141, opened 2019-07-02
Script-based builds

anpaz-msft:
We're moving all our repos to use some well-defined scripts for the build steps so we can rely on those instead of .yaml files when doing cross-repo builds. These changes implement such scripts and updates the yaml files to use them accordingly.

I'm also using this to add the `validate-notebooks.ps1` into the actual, so Jupyter notebooks get validated during build.
Some notebooks still fail during the execution, as they require further fixes on the Katas package; for now, I'm just blacklisting them until we get to fix them.


anpaz-msft:
Oh, and set the build to run in all platforms, to make sure people can run these in the computer they enjoy. :)

scripts/validate-notebooks.ps1
```diff
@@ -1,51 +1,67 @@
...
 
-# List of Katas to verify:
+# List of Notebooks that can't be validated as they required user-input
+# other than the kata answer:
+$not_ready = 
+@(
+    'Check.ipynb',
+    'CHSHGame.ipynb',
+    'DeutschJozsaAlgorithm.ipynb',
+    'GHZGame.ipynb',
```
tcNickolas: What are the conditions that put a kata in this list? Looking at `GHZGame`, it does not seem to have anything odd going on with it (like open-ended tasks or using a different simulator). Is it that one of the tests depends on solution to one of the previous tasks? That should have been in https://github.com/microsoft/QuantumKatas/pull/116.

It would be really nice to document this a bit.

scripts/steps.yml
```diff
@@ -0,0 +1,40 @@
...
+
+- script: pip install setuptools wheel pytest jupyter
+  displayName: 'Install Python tools'
+
+- powershell: ./install-iqsharp.ps1
+  displayName: "Installing IQ#"
+  workingDirectory: $(System.DefaultWorkingDirectory)/scripts
+
+##
+# Build, test & pack
```
tcNickolas: Is there a way to do the building and (most importantly!) the testing locally using the same tools? Could you please document this?

anpaz-msft: Yes, it's very easy: just run the `scripts\build.ps1` and then run the `scripts\test.ps1` from powershell.
You will need to have installed the dependencies (powershell, dotnet, juypter).
I don't have time to document it right now, though.

scripts/validate-notebooks.ps1
anpaz-msft: The condition was: did it ran successfully? In general, I didn't took the time to figure out why it failed. I know that for DJ the problem is that simulate uses an operation defined in a Kata, which is currently not supported (we would need to do a similar trick to what we did for the games), for the others I suspect something similar.

scripts/.gitignore
```diff
@@ -0,0 +1 @@
+.tools
```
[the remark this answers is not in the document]
anpaz-msft: It could, but it is best practice to keep .gitignore changes limited to the place it applies.

scripts/validate-notebooks.ps1
tcNickolas: I spent some time investigating, fixed a couple of katas and documented the ones I couldn't fix in https://github.com/microsoft/QuantumKatas/pull/159.

scripts/steps.yml
tcNickolas: Documented a bit in https://github.com/microsoft/QuantumKatas/pull/159.

Issue #94, opened 2019-04-17
Quantum counting


friguzzi:
Dear all,
I wrote a quantum counting kata but I'm encountering problems, it works for some functions and doesn't for others, should I submit a pull request?

Best
Fabrizio

tcNickolas:
Sure, if you need help figuring out how to make the kata work, you can open a pull request, describe what works and what doesn't work, and we can discuss it. I might not answer very fast (and I apologize for that in advance), as I'm working on bringing in the currently open PRs and on porting some of the katas to Jupyter Notebook format, but I'll try to not disappear altogether :-)

friguzzi:
Hi Mariia,
before submitting the kata, I would like to get the implmentation right.
I have this program
https://gist.github.com/friguzzi/84fbba72f7b6ba7bd58b0c4f8a8be60f
https://gist.github.com/friguzzi/3c7b836d572c26450efbdc2380cbc97e
for counting the solutions of the 3 bits formula
(not reg[0] or reg[2]) and (not reg[1] or reg[2]) and (not reg[0] or not reg[1])
This formula has 4 solutions.
I consider an extended formula with 7 bits that is true when the first three bit satisfy the above formula and the other three bits are at 0. The extended formula this has 4 solutions as well but i can increase easily the number of solutions by excluding variables from the formula.
I compute the phase \phi with the linked program, then I compute \theta/2 as 2*PI*\phi/2=PI*\phi
Now the number of solutions should be 128*sin(\theta/2)^2 right?
I get the correct number of solutions if I compute 128*(1-sin(\theta/2)^2).
It's as if the formula is negated but I can't find where the negation is.
This same formula with GroverSearch returns correct solutions.

Any help is appreciated.

best
Fabrizio
The linked program

tcNickolas:
Sorry for the delay replying, I'll need to dig deeper into this topic to answer, and so far I haven't had time for it. I'll try to figure it out in the next couple of days.

friguzzi:
Hi Mariia,
have you had the chance to have a look at the code?

tcNickolas:
I spent a couple of hours today looking into the algorithm for quantum counting and your code for it - so far everything you do looks reasonable. Why the answer diverges from the expected one, I don't know yet - I'll have to keep looking. Sorry it's taking me so long!

friguzzi:
Thanks Mariia

Benjamin-L:
[...]

PR #147, opened 2019-07-15
BasicGates kata: Improve test harness

tcNickolas:
As was pointed out in #146, task solutions which add extra global phase can be confusing for the learner. This change modifies the test harnesses for part I to check that the solutions do exactly the transformation required and to fail if the solution adds a global phase.

### DIFF
--- a/BasicGates/BasicGates.ipynb
+++ b/BasicGates/BasicGates.ipynb
@@ -73,7 +73,12 @@
     "\n",
     "### Q# materials\n",
     "\n",
-    "* Basic gates provided in Q# belong to the `Microsoft.Quantum.Intrinsic` namespace and are listed [here](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.intrinsic)."
+    "* Basic gates provided in Q# belong to the `Microsoft.Quantum.Intrinsic` namespace and are listed [here](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.intrinsic).\n",
+    "\n",
+    "> Note that all operations in this section have `is Adj+Ctl` in their signature.\n",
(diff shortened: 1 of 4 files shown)


cgranade:
I'm not sure I entirely agree on this point? One of the strengths of Q# thus far has been its commitment to operationalism, in that that only differences which are physically observable are then observable in Q#. Taking that philosophy, two feasible solutions which differ by a global phase are identical solutions. 

Checking the action of each solution under the `Controlled` functor then adds an additional constraint not currently specified in the task description (namely that the task solution is controllable and applies the prescribed basis transformation when controlled), such that I would think we'd need to at least modify the description accordingly.

tcNickolas:
The ability to ignore the global phase *depending on the situation* is a deep philosophical concept, which is beyond the scope of this kata. On the other hand, quietly and habitually ignoring the global phase 1) is very confusing for a beginner who reads the solution for the task and tries to figure out why the transformation done by the solution differs from what the task was asking for, as indicated in https://github.com/microsoft/QuantumKatas/pull/146, and 2) can cause *very* interesting troubleshooting sessions like https://github.com/microsoft/QuantumKatas/issues/94. 

I plan to create a tutorial addressing specifically the nuances of working with the phase; however, in this kata (which for most solvers is the very first attempt to try out quantum computing gates) it is better to check that the transformation that passes the tests is exactly the same as the transformation required.

I will add some language addressing the `+Ctl` part. It's not a heavy burden, given that at this point people only use intrinsic gates, so it is very hard to come up with something that would not be automatically controllable; on the other hand, if the kata starts with a requirement to write your code in a way that allows to generate adjoint and controlled variants automatically, it has to explain what these are - and this will get very scary very fast.

Issue #68, opened 2019-02-22
Measurement Task 1.7 - code passes even with more than one measurement


GaurangTandon:
The comment description for that task reads:

    // You can use exactly one measurement.

The following code passes:

    operation TwoBitstringsMeasurement (qs : Qubit[], bits1 : Bool[], bits2 : Bool[]) : Int {
        let n = Length(qs);

        for(i in 0..n-1){
            if(M(qs[i]) == One){
                if(bits1[i] && bits2[i]){
                }elif(bits1[i]){
                    return 0;
                }else{
                    return 1;
                }
            }else{
                if(not bits1[i] && not bits2[i]){
                }elif(not bits1[i]){
                    return 0;
                }else{
                    return 1;
                }
            }
        }

        return -1;
    }

AFAICT, the above code makes n-1 measurements, if both the boolean bit arrays have different value only at the last bit.

tcNickolas:
Good point, this is one of the older katas and the testing framework doesn't count the operations used, like the later (and more sophisticated) katas do. I'll need to revisit this at some point and see whether the rest of the tasks in the Measurements kata need to have more strict checks as well.

Thank you for pointing this out!

tcNickolas:
This has been resolved as part of #101.

PR #155, opened 2019-07-23
QEC_BitFlipCode: Add check for the number of measurements used in task 1

tcNickolas:
This fixes an issue similar to #68: the task requires to use only one measurement but doesn't enforce this requirement in the testing harness.

### DIFF
--- a/QEC_BitFlipCode/Tests.qs
+++ b/QEC_BitFlipCode/Tests.qs
@@ -98,14 +98,14 @@ namespace Quantum.Kata.QEC_BitFlipCode {
             let res = MeasureParity(register);
             
             // check that the returned parity is correct
-            EqualityFactB(res == Zero, parity == 0, $"Failed on {stateStr}.");
+            Fact((res == Zero) == (parity == 0), $"Failed on {stateStr}.");
             
             // check that the state has not been modified
             Adjoint statePrep(register);
             AssertAllZero(register);
 
             let nm = GetOracleCallsCount(M) + GetOracleCallsCount(Measure);
-            AssertBoolEqual(nm <= 1, true, $"You are allowed to do at most one measurement, and you did {nm}");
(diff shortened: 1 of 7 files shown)


QEC_BitFlipCode/Tests.qs
```diff
@@ -91,14 +92,20 @@ namespace Quantum.Kata.QEC_BitFlipCode {
...
             
             // check that the returned parity is correct
-            EqualityFactB(res == Zero, parity == 0, $"Failed on {stateStr}.");
+            Fact((res == Zero) == (parity == 0), $"Failed on {stateStr}.");
             
             // check that the state has not been modified
             Adjoint statePrep(register);
             AssertAllZero(register);
+
+            let nm = GetOracleCallsCount(M) + GetOracleCallsCount(Measure);
```
cgranade: That's odd, does CounterSimulator not catch calls from within `M`? In [`Intrinsic.qs`](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Intrinsic/Intrinsic.qs#L724), `M` is implemented by a call to `Measure`, such that I'd naïvely thought this would double-count.

tcNickolas: I just double-checked that one call to `M` yields 1 count of `M` and 0 counts of `Measure`, so something is working differently than you expect. `MResetZ` also counts as 1 `M` and 0 `Measure`.

alan-geller: CounterSimulator derives from Quantum Simulator, which provides an intrinsic implementation of `M` that overrides the implementation in `Instrinsic.qs`.

tcNickolas: Thank you for the explanation, this inconsistency was worrying me :-)

cgranade: Ah, of course. I forgot that `QuantumSimulator` has an implementation for both `M` and `Measure`, sorry. Thanks!

Issue #74, opened 2019-02-26
JointMeasurements - Task 7 - invalid code passes


GaurangTandon:
```
// Task 7**. Controlled X gate with arbitrary target
// Input: Two qubits (stored in an array of length 2) in an arbitrary
//        two-qubit state α|00⟩ + β|01⟩ + γ|10⟩ + δ|11⟩.
// Goal:  Change the two-qubit state to α|00⟩ + β|01⟩ + δ|10⟩ + γ|11⟩ using only single-qubit gates and joint measurements.
//        Do not use two-qubit gates.
operation ControlledX_General (qs : Qubit[]) : Unit {

    body (...) {
        CNOT(qs[0], qs[1]);
    }

    adjoint self;
}
```

The above code passes, even though it is written in the docstring that "Do not use two-qubit gates".

tcNickolas:
This is similar to #68 - test harness is not sophisticated enough to validate this kind of restrictions. It is a little tricky to validate - it is very easy to check that the code doesn't use `CNOT` or `Controlled X` gates, but `CNOT` can be expressed using a wide variety of other gates (`CCNOT` or `Controlled Z` ), which means that we need indeed to prohibit any 2-qubit gates.

I'll need to think about it. Meanwhile, treat this as a helpful suggestion to direct the learning process - if a task asks to to implement `CNOT` in a certain way, you can of course just use the `CNOT` to pass the test, but you've hardly learned anything this way :-)

tcNickolas:
#170 finally fixes this issue.

PR #158, opened 2019-07-24
CounterSimulator: Add ability to count multi-qubit operations

tcNickolas:
In some tasks, we need to restrict usage of certain multi-qubit operations. To do this, `CounterSimulator` needs to be able to track the number of qubit parameters the operation received.

This will be part of the fix for #74, among its other uses. The second part of the fix will come after we update the Nuget package with this change and enable the kata Notebooks to use it.


tcNickolas:
The code in `JointMeasurements` kata that exercises the new feature looks like this:

    // A pair of operations to fine-tune universal CounterSimulator
    // for the purposes of the last two tasks: prohibiting all multi-qubit operations,
    // except the two that can be used for solving this task
    operation ResetMultiQubitNonMeasurementOpCount () : Unit {
        ResetMultiQubitOpCount();
        ResetOracleCallsCount();
    }

    operation GetMultiQubitNonMeasurementOpCount () : Int {
        return GetMultiQubitOpCount() - GetOracleCallsCount(Measure) - GetOracleCallsCount(MeasureAllZ);
    }

Then I require that `GetMultiQubitNonMeasurementOpCount() <= 1` in the tasks that prohibit use of non-measurement multiqubit operations. The `1` accounts for calling the task that is being tested itself.

utilities/Common/CounterSimulator.cs
```diff
@@ -193,5 +205,60 @@ public GetMaxQubitCountImpl(CounterSimulator m) : base(m)
...
+        #region Count multi-qubit operations (except Measure)
+        public void CountMultiQubitOperations(ICallable op, IApplyData data)
+        {
+            int nQubits = 0;
+            using (IEnumerator<Qubit> enumerator = data?.Qubits?.GetEnumerator()) {
+                if (enumerator is null) {
+                    // the operation doesn't have qubit parameters
+                    return;
+                }
+                while (enumerator.MoveNext() && nQubits < 2)
```
anpaz-msft: I would do `nQubits = enumerator.Take(2).Count();`.

utilities/Common/CounterSimulator.cs
```diff
@@ -193,5 +205,60 @@ public GetMaxQubitCountImpl(CounterSimulator m) : base(m)
...
+            }
+            if (nQubits >= 2) {
+                _multiQubitOperations++;
+            }
+        }
+
+        /// <summary>
+        /// Custom operation to reset the numbers of multi-qubit operations.
+        /// </summary>
+        public class ResetMultiQubitOpCountImpl : ResetMultiQubitOpCount
```
anpaz-msft: If you go the always count multi-qubit gates, this is probably not needed.

utilities/Common/CounterSimulator.cs
tcNickolas: `Take` seems to be a method of LINQ `IEnumerable`, while `enumerator` is `IEnumerator`, and I can't find a neat way to cast `IEnumerator` to `IEnumerable`. I'm not sure the benefits of using `.Take` outweigh the hassle of adding LINQ to the dependencies and casting the enumerables.

PR #160, opened 2019-08-01
Update SuperdenseCoding.ipynb

vivienlonde:
changed two signs in the hint of Task 2.


vivienlonde:
changed two signs in the hint of Task 2.

PR #161, opened 2019-08-06
Add task for even numbers superposition

evilpegasus:
Added task 10 to superposition


Superposition/ReferenceImplementation.qs
```diff
@@ -175,21 +185,6 @@ namespace Quantum.Kata.Superposition {
         }
     }
 
-    // ------------------------------------------------------
```
tcNickolas: The whitespace coming back and this task disappearing look like you've created your branch off the master earlier than https://github.com/microsoft/QuantumKatas/commit/debff12623dd85fc4dd2651b0b14c02a0c585a0d. Could you please make sure that your changes are based on the latest state of master branch and don't undo any other changes?

Superposition/ReferenceImplementation.qs
```diff
@@ -115,45 +115,55 @@ namespace Quantum.Kata.Superposition {
...
     operation AllBasisVectorsSuperposition_Reference (qs : Qubit[]) : Unit is Adj {
-
+        
         for (q in qs) {
             H(q);
         }
     }
 
-
+	// Task 10. Superposition of all even numbers
```
tcNickolas: Please use 4 spaces instead of tabs for indentation to be consistent with the rest of the file. (The same applies to the changes in `Tasks.qs` and `Tests.qs`)

Superposition/ReferenceImplementation.qs
```diff
@@ -115,45 +115,55 @@ namespace Quantum.Kata.Superposition {
...
+    // Input: N qubits in |0...0⟩ state.
+	// Goal: create a superposition of all even numbers on n qubits
```
tcNickolas:
```suggestion
	// Goal: create a superposition of all even numbers on N qubits.
```

Superposition/Tasks.qs
```diff
@@ -118,31 +118,34 @@ namespace Quantum.Kata.Superposition {
...
     // Goal: create an equal superposition of all basis vectors from |0...0⟩ to |1...1⟩
     // (i.e. state (|0...0⟩ + ... + |1...1⟩) / sqrt(2^N) ).
     operation AllBasisVectorsSuperposition (qs : Qubit[]) : Unit {
         // ...
     }
+    
+    
+	// Task 10. Superposition of all even numbers
+    // Input: N qubits in |0...0⟩ state.
+	// Goal: create a superposition of all even numbers on n qubits
```
tcNickolas:
```suggestion
	// Goal: create a superposition of all even numbers on N qubits.
```

Superposition/Tasks.qs
```diff
@@ -118,31 +118,34 @@ namespace Quantum.Kata.Superposition {
...
+	operation EvenNumbersSuperposition (qs : Qubit[]) : Unit is Adj {
```
tcNickolas: Since this is the first time the kata refers to numbers instead of just basis states, you need to provide more details on how the numbers are encoded as basis states. I've found the endian-ness to be a particular challenge, so we'd have to say something like "A basis state represents a number using big-endian binary notation; for example, a state |01⟩ corresponds to an integer 1, and a state |10⟩ - to an integer 2."

Superposition/Superposition.ipynb
```diff
@@ -367,37 +367,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Task 11*. Hardy state.\n",
```
tcNickolas: Please add the new task to the Notebook as well, so that everybody who attempts to solve the kata can enjoy it regardless of the format they use :-)

Superposition/Tasks.qs
tcNickolas: I would suggest adding a little twist to this task to make it more interesting: add an extra parameter `odd : Bool` and specify that if it is true you need to prepare a superposition of all odd numbers, and if it is false - of all even numbers. The difference in the reference implementation will be just applying an X gate to qs[N-1] if odd is true (which corresponds to flipping the least significant bit), but it will make it interestingly different from the previous task.

Superposition/Tasks.qs
```diff
@@ -129,21 +129,31 @@ namespace Quantum.Kata.Superposition {
     }
 
 
-    // Task 10. |00⟩ + |01⟩ + |10⟩ state
+    // Task 10. Superposition of all even numbers
+    // Input: N qubits in |0...0⟩ state.
+	// Goal: create a superposition of all even numbers on N qubits.
```
tcNickolas: This line and lines 135 and 137 still have tabs in them (you can see GitHub indents them the same as line 136 even though 136 should be the next level of indentation)

Superposition/Superposition.ipynb
```diff
@@ -560,10 +584,16 @@
    "name": "iqsharp"
   },
   "language_info": {
-   "file_extension": ".qs",
```
tcNickolas: Looks like something weird happened to the Notebook - language info should keep indicating Q#, not switch to Python. Did you open and edit it as Q# Notebook, or did you run into any trouble doing that? Continuous integration also indicates that the file is no longer a properly formatted JSON file. Could you please look into that?

evilpegasus:
TODO:
Add check for odd numbers
Write better description for task
Give more info about notation (hints)

tcNickolas:
This looks good, all the right pieces are in the right places! Once you finish the list of TODOs you have, it will be ready to be merged.

One note, I think the little-endian notation is better introduced in task description and not in a hint; hints are supposed to push the solver in the right direction of though, and the endianness is really part of the task description.

PR #167, opened 2019-08-16
Superposition measurement

melmatt7:


Measurements/Tasks.qs
```diff
@@ -126,9 +127,29 @@ namespace Quantum.Kata.Measurements {
         // ...
         return -1;
     }
-    
-    
-    // Task 1.8. |0...0⟩ state or W state ?
+
+// Task 1.8. Distinguish two superposition states given by two arrays of bit strings
```
tcNickolas:
```suggestion
    // Task 1.8. Distinguish two superposition states given by two arrays of bit strings
```

Measurements/Tasks.qs
```diff
@@ -126,9 +127,29 @@ namespace Quantum.Kata.Measurements {
...
+    // Inputs:
+    //      1) N qubits in |0...0⟩ state which are guaranteed to be
+    //         in one of the two superposition states described by the given bit strings.
+    //      2) two arrays of bit strings represented as bit strings containing Bool[]s.
+    // Output: 0 if qubits were in the basis state(s) described by the first superposition bit string,
+    //         1 if they were in the basis state(s) described by the second superposition bit string.
+    // Bit values false and true correspond to |0⟩ and |1⟩ states.
+    // The state of the qubits at the end of the operation does not matter.
+    // You are guaranteed that both the arrays of bit strings will have the same length
```
tcNickolas: This constraints seems unnecessary? It doesn't look like distinguishing superpositions of several basis states from one another should be dependent on the exact number of terms in the superpositions - I can imagine wanting to distinguish 00 from 10 + 11

Measurements/Tasks.qs
```diff
@@ -126,9 +127,29 @@ namespace Quantum.Kata.Measurements {
...
+    // of either 1, 2 or 4.
+    // You are also guaranteed that each bitstring within each array has the same length.
+    // All bit strings, between both arrays, will differ in at least one bit.
```
tcNickolas: I'm not quite sure what exactly does this constraint mean (on this line and the next one) and what is the expected solution, sorry. 

If it means that there is at least one position in which all bit strings in the first array have a certain value and all bit strings in the second array have another value (in which case you can indeed distinguish them in exactly one measurement), this is a good extension of the previous task, but you need:

* to say so in more detail in the task description,
* to fix the example (the current example doesn't have one bit that you can measure to distinguish the superpositions with 100% guarantee), 
* to fix the reference solution to find the differing bit before any measurements are done (the current solution can do multiple measurements before it gets to the location of that differing bit),
* and to enforce the max number of measurements to be 1 by passing `measurementsPerRun = 1` to the test method `DistinguishStates_MultiQubit`

If it is a different task, in which indeed all the bit strings need to differ from each other with no other constraints, then the reference solution needs to be reworked as well.

Sorry I'm not posting these comments in the relevant places, they would be a pain to track individually!

Measurements/ReferenceImplementation.qs
```diff
@@ -115,18 +117,75 @@ namespace Quantum.Kata.Measurements {
...
+    // The state of the qubits at the end of the operation does not matter.
+    // You are guaranteed that both the arrays of bit strings will have the same length
+    // of either 1, 2 or 4.
+    // You are also guaranteed that each bitstring within each array has the same length.
+    // All bit strings, between both arrays, will differ in at least one bit.
+    // You can use exactly one measurement.
+    // Example: for bit string arrays [[false, true, false], [false, false, true]] and [[true, true, true], [false, true, true]]
+    //          return 0 corresponds to state |010⟩ or |001⟩, and return 1 corresponds to state |111⟩ or |011⟩.
+
+    function CheckSuperposition_Reference (superpositionArray : Bool[][], res : Bool, i : Int) : Int {
```
tcNickolas: Could you please clarify what is this function supposed to do? It looks like it's counting the number of bit strings that do *not* have res in their i-th position, but I'm pretty sure this can be done much easier and faster, without a mutable array. It can also be modified to check whether all bit strings have res (or not res) in their i-th position, in which case you'll be able to return early, as soon as you find a mismatch (or a match).

PR #170, opened 2019-09-08
JointMeasurements: Restrict the use of multi-qubit gates in tasks 6 and 7

tcNickolas:
This is the second part of the fix for #74.


tcNickolas:
Yes, the new construct can be used in a fair number of tests across multiple katas (and probably in some solutions as well). I prefer to keep the changes with separate logic in separate PRs to simplify tracking them later. 

("Using new features of Q#" can be a nice "good first issue" work item, btw)

PR #172, opened 2019-09-17
1.4 sign fix

peterse:
The accepted answer to this has a coefficient of `2.0 * \alpha` but the signs given in the description correspond to an angle of `-2.0 * \alpha` according to the Ry doc at https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.ry?view=qsharp-preview


tcNickolas:
Could you please elaborate on how you got these signs? Ry matrix is the following:

image

Applying this gate to state |0⟩ will produce the first column of this matrix (you have to multiply the matrix by the column vector (1, 0)ᵀ, and you'll get column vector (cos α, sin α)ᵀ which translates to cos α |0⟩ + sin α |1⟩).

You can also use DumpMachine to check the signs of the amplitudes you get after transformation using the following code. 

    // Allocate a qubit which you'll pass to the operation
    using (q = Qubit()) {
        // The qubit starts in |0⟩ state; if you need a different state, apply the gates to prepare it

        // Call the operation you're testing
        Ry(2.0 * PI() * 0.25, q);

        // You can print the amplitudes of the wave function to a file using function DumpMachine
        // from namespace Microsoft.Quantum.Diagnostics
        DumpMachine("C://Tmp//task1.txt");

        // Remember to return qubits to state |0⟩ before releasing them
        Reset(q);
    }

peterse:
on second thought it seems I am the one who made the error. Sorry!

tcNickolas:
No worries, it's better to double-check than to miss something!

PR #177, opened 2019-10-02
task 1.7 global phase change was added

sdf94:


BasicGates/Tests.qs
```diff
@@ -119,22 +125,23 @@ namespace Quantum.Kata.BasicGates {
...
     // ------------------------------------------------------
-    operation T17_BellStateChange1_Test () : Unit {
+    operation T18_BellStateChange1_Test () : Unit {
         VerifyBellStateConversion(BellStateChange1, 0, 1);
     }
     
     
     // ------------------------------------------------------
-    operation T18_BellStateChange2_Test () : Unit {
+    operation T19_BellStateChange2_Test () : Unit {
```
tcNickolas: The `T...` prefix was introduced to order the tests when they show up in Visual Studio Test panel (or in VS Code). Initially there were less than 10 tasks in the first part, so the prefix was T - part number - 1-digit task number. The task you're adding is the 10th one, so the `T110_` task will appear immediately after `T11_`, which is not as neat as it could be.

Could you please go over the test operations in this file (and in the matching `%kata` entries in the notebook) and rename them to have 2-digit task numbers? For example,
```suggestion
    operation T109_BellStateChange2_Test () : Unit {
```
Thank you!

BasicGates/Tasks.qs
```diff
@@ -111,24 +111,34 @@ namespace Quantum.Kata.BasicGates {
         // ...
     }
 
+    // Task 1.7. Global Phase Change
+    // Inputs:
+    //     1) Pauli operator (μ) represented as Pauli.
+    //     2) Angle alpha, in radians, represented as Double.
+    //     3) A qubit in state β|0⟩ + γ|1⟩.
+    // Goal: Change the state of the qubit to - β|0⟩ - γ|1⟩.
+    operation GlobalPhaseChange(pauli: Pauli, alpha: Double, q: Qubit) : Unit is Adj+Ctl{
```
tcNickolas: If the task is just to add a `-1` global phase to the qubit state, you don't need to pass the first two of these parameters - just the qubit itself. 
```suggestion
    operation GlobalPhaseChange (q: Qubit) : Unit is Adj + Ctl {
```
This way the reference solution stays the same, `R(PauliI, 2.0 * PI(), q);`, but there are fewer parameters to deal with (the way it is phrased now, the purpose of these parameters is unclear, and you only test the solution with one value). 

If you're worried that people will not be able to find `PI()` function, you can add a hint telling them the namespace they need to open for it (they'll need to use the constant with fairly high precision - I've seen solutions that take 3.1415 fail being not accurate enough).

In addition, this change allows people to write other solutions - for example, ZXZX will add a -1 phase without any rotations and constants required.

BasicGates/BasicGates.ipynb
```diff
@@ -264,7 +264,37 @@
...
-    "### Task 1.7. Bell state change - 1\n",
+    "### Task 1.7. Global Phase Change",
+    "\n",
+    "**Input:** \n",
+    "\n",
+    "1) Pauli operator (μ) represented as Pauli.\n",
+    "\n",
+    "2) Angle alpha, in radians, represented as Double.\n",
+    "\n",
+    "3) A qubit in state β|0⟩ + γ|1⟩.\n",
```
tcNickolas: In notebook version of the kata we usually write qubit states as formulas - they look neater that way

```suggestion
    "3) A qubit in state $|\\psi\\rangle = \\beta |0\\rangle + \\gamma |1\\rangle$.\n",
```

The same applies to the state in the goal.

tcNickolas:
(I also took a look at the CI failure - it is a transient error I've been seeing sometimes, nothing to worry about in relation to your change)

BasicGates/BasicGates.ipynb
sdf94: I was wondering about this! Thanks for the clarification.

BasicGates/Tasks.qs
```diff
@@ -111,24 +111,34 @@ namespace Quantum.Kata.BasicGates {
         // ...
     }
 
+    // Task 1.7. Global Phase Change
+    // Inputs:
+    //     1) Pauli operator (μ) represented as Pauli.
+    //     2) Angle alpha, in radians, represented as Double.
+    //     3) A qubit in state β|0⟩ + γ|1⟩.
+    // Goal: Change the state of the qubit to - β|0⟩ - γ|1⟩.
+    operation GlobalPhaseChange (q: Qubit) : Unit is Adj + Ctl {
```
tcNickolas: Since you changed the signature of this operation, you need to do matching changes in other parts of the code:
* the list of inputs on lines 116-118 of this file
* the signature and list of inputs in `ReferenceImplementation.qs` (and the reference implementation itself, to use 2.0 * PI() instead of the parameter)
* the signature of the matching operation in `BasicGates.ipynb`
* the invocation of the solution and the reference implementation in `Tests.qs` (I think that's what it causing the build break in continuous integration)

sdf94: I am not sure if this is how the function is suppose to look like in Tests.qs

tcNickolas: Yes, the syntax could be a bit more concise (operation name with (_) as an argument can be written as just operation name, since there is no partial application necessary in this case), but the idea is right

PR #179, opened 2019-10-03
Updating projects to 0.9.1909.3002

anpaz-msft:


anpaz-msft:
Yes it should, good catch. The update script was not taking this into consideration, I've added it.

PR #182, opened 2019-10-05
Linear algebra

flockofonions:
Added a tutorial teaching Linear Algebra


tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "* Complex matrices"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook has several tasks that require you to write Python code to test your understanding of the concepts. If you are not familiar with Python, [here](https://docs.python.org/3/tutorial/index.html) is a good introductory tutorial for it.\n",
+    "\n",
+    "The exercises use Python's built-in representation of complex numbers. Most of the operations (addition, multiplication, etc.) work as you expect them to, but there are a few Python-specific caveats:\n",
```
tcNickolas:
```suggestion
    "The exercises use Python's built-in representation of complex numbers. Most of the operations (addition, multiplication, etc.) work as you expect them to. Here are a few notes on Python-specific syntax:\n",
```

tcNickolas: I would make the section about complex numbers (from this sentence up to "You can find more information...") a note (marked with `>`) to split the wall of text into several sections.

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Part I. Matrices and Basic Operations\n",
+    "\n",
+    "## Matrices and Vectors\n",
+    "\n",
+    "A **matrix** is a rather simple concept - it's a set of numbers grouped together in a rectangular grid. Here is a $2$ by $2$ matrix:\n",
```
tcNickolas:
```suggestion
    "A **matrix** is set of numbers arranged in a rectangular grid. Here is a $2$ by $2$ matrix:\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "\n",
+    "$$A =\n",
+    "\\begin{bmatrix} 1 & 2 \\\\ 3 & 4 \\end{bmatrix}$$\n",
+    "\n",
+    "$A_{i,j}$ Refers to the element in row $i$ and column $j$ of matrix $A$. In the above example, $A_{0,1} = 2$.\n",
```
tcNickolas:
```suggestion
    "$A_{i,j}$ refers to the element in row $i$ and column $j$ of matrix $A$ (all indices are 0-based). In the above example, $A_{0,1} = 2$.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "\n",
+    "An $n \\times m$ matrix will have $n$ rows and $m$ columns, like so:\n",
+    "\n",
+    "$$\\begin{bmatrix}\n",
+    "    x_{0,0} & x_{0,1} & \\dotsb & x_{0,m-1} \\\\\n",
+    "    x_{1,0} & x_{1,1} & \\dotsb & x_{1,m-1} \\\\\n",
+    "    \\vdots  & \\vdots  & \\ddots & \\vdots  \\\\\n",
+    "    x_{n-1,0} & x_{n-1,2} & \\dotsb & x_{n-1,m-1}\n",
```
tcNickolas:
```suggestion
    "    x_{n-1,0} & x_{n-1,1} & \\dotsb & x_{n-1,m-1}\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "\\end{bmatrix}$$\n",
+    "\n",
+    "A $1 \\times 1$ matrix is equivalent to a scalar:\n",
+    "\n",
+    "$$\\begin{bmatrix} 3 \\end{bmatrix} = 3$$\n",
+    "\n",
+    "In quantum computing, we use complex valued matrices: the elements of a matrix can be complex numbers. This, for example, is a valid complex-valued matrix:\n",
```
tcNickolas:
```suggestion
    "Quantum computing uses complex-valued matrices: the elements of a matrix can be complex numbers. This, for example, is a valid complex-valued matrix:\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "\n",
+    "$$\\begin{bmatrix}\n",
+    "    1 & i \\\\\n",
+    "    -2i & 3 + 4i\n",
+    "\\end{bmatrix}$$\n",
+    "\n",
+    "Finally, for the purposes of linear algebra, a **vector** is an $n \\times 1$ matrix. Here, for example, is a $3 \\times 1$ vector:\n",
```
tcNickolas:
```suggestion
    "Finally, a **vector** is an $n \\times 1$ matrix. Here, for example, is a $3 \\times 1$ vector:\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "$$\\begin{bmatrix} 1 \\\\ 2i \\\\ 3 + 4i \\end{bmatrix}$$"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Matrix Addition\n",
+    "\n",
+    "The easiest matrix operation to understand is **matrix addition**. Matrix addition works between two matrices of the same size, and adds each number from the first matrix to the number in the same position in the second matrix:\n",
```
tcNickolas:
```suggestion
    "The easiest matrix operation is **matrix addition**. Matrix addition works between two matrices of the same size, and adds each number from the first matrix to the number in the same position in the second matrix:\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "    x_{0,0} + y_{0,0} & x_{0,1} + y_{0,1} & \\dotsb & x_{0,m-1} + y_{0,m-1} \\\\\n",
+    "    x_{1,0} + y_{1,0} & x_{1,1} + y_{1,1} & \\dotsb & x_{1,m-1} + y_{1,m-1} \\\\\n",
+    "    \\vdots  & \\vdots  & \\ddots & \\vdots  \\\\\n",
+    "    x_{n-1,0} + y_{n-1,0} & x_{n-1,1} + y_{n-1,1} & \\dotsb & x_{n-1,m-1} + y_{n-1,m-1}\n",
+    "\\end{bmatrix}$$\n",
+    "\n",
+    "Matrix addition has the following properties:\n",
+    "\n",
+    "* Commutativity: $A + B = B + A$\n",
+    "* Associativity: $(A + B) + C = A + (B + C)$"
```
tcNickolas: I wonder if we should mention matrix subtraction somewhere? No need to write out the formulas, something like this: "Similarly, we can subtract one matrix from another: A − B is computed by subtracting elements of B from corresponding elements of A."

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "    \n",
+    "    # You can use this function to initialize a matrix filled with 0s\n",
+    "    c = create_empty_matrix(n, m)\n",
+    "    \n",
+    "    # You can use a for loop to do something n (or m) times\n",
+    "    for i in range(n):\n",
+    "        # Loops can be nested\n",
+    "        for j in range(m):\n",
+    "            # You can access elements of a matrix like this\n",
+    "            x = a[i][j]\n",
```
tcNickolas: Following the approach we used in CompexArithmetic, I would add a line `y = b[i][j]` here, so that the learner has to modify just the line marked with `...`

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "    (1 \\cdot 1) + (2 \\cdot 3) + (3 \\cdot 5) & (1 \\cdot 2) + (2 \\cdot 4) + (3 \\cdot 6) \\\\\n",
+    "    (4 \\cdot 1) + (5 \\cdot 3) + (6 \\cdot 5) & (4 \\cdot 2) + (5 \\cdot 4) + (6 \\cdot 6)\n",
+    "\\end{bmatrix}\n",
+    "=\n",
+    "\\begin{bmatrix}\n",
+    "    22 & 28 \\\\\n",
+    "    49 & 64\n",
+    "\\end{bmatrix}$$\n",
+    "\n",
+    "Note that matrix multiplication is **not commutative:** $AB$ rarely equals $BA$. That being said, here are the properties it does have:\n",
```
tcNickolas: It seems better to start with the properties that is does have: 
```suggestion
    "Matrix multiplication has the following properties:\n",
```
(and make "Note that matrix multiplication is not commutative: 𝐴𝐵 rarely equals 𝐵𝐴." the last entry in the list of properties)

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "* Distributivity over matrix addition: $A(B + C) = AB + AC$ and $(A + B)C = AC + BC$\n",
+    "* Associativity with scalar multiplication: $xAB = x(AB) = A(xB)$\n",
+    "\n",
+    "Another very important property of matrix multiplication is that a matrix multiplied by a vector produces another vector.\n",
+    "\n",
+    "An **identity matrix** $I_n$ is a special $n \\times n$ matrix which has $1$s on the main diagonal, and $0$s everywhere else:\n",
+    "\n",
+    "$$I_n =\n",
+    "\\begin{bmatrix}\n",
+    "    1 & 0 & 0 & \\dotsb & 0 \\\\\n",
```
tcNickolas: I'd remove one row and one column from the matrix, it seems enough to have just three rows/columns to get the pattern.

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "    return # ..."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Matrix Multiplication\n",
+    "\n",
+    "A more complicated, but much more important and interesting operation is **matrix multiplication**. The first complicated thing about it is that neither its operands nor its output are the same size: an $n \\times m$ matrix multiplied by an $m \\times k$ matrix results in an $n \\times k$ matrix. That is, for matrix multiplication to be applicable, the number of columns in the first matrix must equal the number of rows in the second matrix.\n",
```
tcNickolas: What is the second complicated thing? Generally I'd try to steer away from the use of the word "complicated" here to avoid scaring people, since it's not really that complicated. Could you rewrite it in some less intimidating terms?

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,984 @@
...
+    "1. An $n \\times m$ matrix $A$.\n",
+    "2. An $m \\times k$ matrix $B$.\n",
+    "\n",
+    "**Goal:** Return the $n \\times k$ matrix that results from $AB$.\n",
+    "\n",
+    "<br/>\n",
+    "<details>\n",
+    "    <summary><strong>Need a hint? Click here</strong></summary>\n",
+    "    To solve this exercise, you will need 3 <code>for</code> loops: one to go over the rows in the output matrix, one to go over the columns, and one to add up the products:\n",
+    "    <pre>\n",
```
tcNickolas: Do you need both `<pre>` and `<code>` tags here? `<code>` on its own looks fine to me (Firefox)

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,984 @@
...
+   "metadata": {},
+   "source": [
+    "### <span style=\"color:blue\">Exercise 3</span>: Matrix multiplication.\n",
+    "\n",
+    "**Inputs:**\n",
+    "\n",
+    "1. An $n \\times m$ matrix $A$.\n",
+    "2. An $m \\times k$ matrix $B$.\n",
+    "\n",
+    "**Goal:** Return the $n \\times k$ matrix that results from $AB$.\n",
```
tcNickolas:
```suggestion
    "**Goal:** Return the $n \\times k$ matrix equal to the matrix product $AB$.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,984 @@
...
+    "<br/>\n",
+    "<details>\n",
+    "    <summary><strong>Need a hint? Click here</strong></summary>\n",
+    "    To solve this exercise, you will need 3 <code>for</code> loops: one to go over the rows in the output matrix, one to go over the columns, and one to add up the products:\n",
+    "    <pre>\n",
+    "        <code>\n",
+    "    for i in range(n):\n",
+    "        for j in range(k):\n",
+    "            sum = 0\n",
+    "            for x in range(m):\n",
```
tcNickolas: I'd use a different variable name for this loop to match the index we used in the formula definition of matrix multiplication:
```suggestion
    "            for t in range(m):\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,986 @@
...
+    "    2\n",
+    "\\end{bmatrix}\n",
+    "=\n",
+    "\\begin{bmatrix}\n",
+    "    4 \\\\\n",
+    "    4 \\\\\n",
+    "    8\n",
+    "\\end{bmatrix}$$\n",
+    "\n",
+    "Notice that the resulting vector is just the initial vector multiplied by a scalar (in this case 4). There is a special name for these values in linear algebra. In general:\n",
```
tcNickolas:
```suggestion
    "Notice that the resulting vector is just the initial vector multiplied by a scalar (in this case 4). This behavior is so noteworthy that it is described using a special set of terms.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,986 @@
...
+    "\n",
+    "Given a nonzero matrix $A$, nonzero vector $V$, and scalar $x$: if $AV = xV$, then $x$ is an **eigenvalue** of $A$, and $V$ is an **eigenvector** of that eigenvalue.\n",
```
tcNickolas:
```suggestion
    "Given a nonzero $n \\times n$ matrix $A$, a nonzero vector $V$, and a scalar $x$, if $AV = xV$, then $x$ is an **eigenvalue** of $A$, and $V$ is an **eigenvector** of $A$ corresponding to that eigenvalue.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,986 @@
...
+    "\n",
+    "Now, of course, this doesn't happen with every combination of matrix, vector, and scalar. Eigenvalues of a matrix are the roots of that matrix's **characteristic polynomial**. An $n \\times n$ matrix $A$'s characteristic polynomial is $|A - (x \\cdot I_n)| = 0$, where $|A|$ is, as you might remember, the matrix's determinant. \n",
```
tcNickolas:
```suggestion
    "Of course, this doesn't happen with every combination of matrix, vector, and scalar. To find eigenvalues of a matrix, we have to consider its **characteristic polynomial** $|A - (x \\cdot I_n)|$, where $| A |$ denotes the determinant of a matrix $A$. The eigenvalues of a matrix are solutions of the equation $|A - (x \\cdot I_n)| = 0$.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
```diff
@@ -0,0 +1,986 @@
...
+   "source": [
+    "### <span style=\"color:blue\">Exercise 14</span>**: Finding an eigenvector.\n",
+    "\n",
+    "**Inputs:**\n",
+    "\n",
+    "1. A $2 \\times 2$ matrix $A$.\n",
+    "2. An eigenvalue $x$ of matrix $A$.\n",
+    "\n",
+    "\n",
+    "**Goal:** Return an eigenvector of $A$ that is associated with $x$.\n",
```
tcNickolas: Let's stress that any eigenvector will be accepted as a solution.
```suggestion
    "**Goal:** Return any eigenvector of $A$ that is associated with $x$.\n",
```

tutorials/LinearAlgebra/LinearAlgebra.ipynb
tcNickolas: Actually, the more I think about this paragraph, the more I doubt that the learner needs it. We don't have a task on finding eigenvalues from scratch, so they won't use this information in the tutorial; we might as well keep it in the external resources. 

Alternatively, we can keep the text and add another exercise for finding eigenvalues of a 2x2 matrix (which ideally would be paired with an exercise for solving a quadratic equation in ComplexArithmetic tutorial). Or we can keep the tutorials as is for now and file an issue to add this pair of tasks later.

What do you think?

tcNickolas:
We also need to modify ComplexArithmetic tutorial as part of this PR to add some closing remarks and a pointer to this tutorial - something like "Congratulations! You've learned everything you need to know about complex numbers are are ready to move on to the next tutorial in the series - linear algebra!". Well, maybe with fewer exclamation points...

tcNickolas:
Most comments show as resolved - is this PR ready for another round of review? Please let me know.

tutorials/LinearAlgebra/LinearAlgebra.ipynb
flockofonions: In Chromium, `<code>` by itself formats the example as several separate segments of code, with thin white lines separating them. `<pre><code>` formats everything as one block.

PR #192, opened 2019-10-08
Add alternate reference implementation to Measurements, problem 1.9

wsgac:
I wrote an alternate reference implementation for `Measurements` problem `1.9`. At slightly higher computational cost it is slightly more readable by treating boolean arrays as numbers.

### DIFF
--- a/Measurements/ReferenceImplementation.qs
+++ b/Measurements/ReferenceImplementation.qs
@@ -194,10 +194,10 @@ namespace Quantum.Kata.Measurements {
     // Alternate reference implementation for task 1.9
     // Slightly more expensive, but uses built-in functions
     operation SuperpositionMeasurement_Alternate (qs : Qubit[], bits1 : Bool[][], bits2 : Bool[][]) : Int {
-        // measure all qubits and, treating the result as a number, check membership in bit arrays
+        // measure all qubits and, treating the result as an integer, check whether it can be found in one of the bit arrays
         let measuredState = ResultArrayAsInt(MultiM(qs));
         for (s in bits1) {
-            if (BoolArrayAsInt(s)==measuredState) {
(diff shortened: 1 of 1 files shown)


Measurements/ReferenceImplementation.qs
```diff
@@ -191,6 +191,20 @@ namespace Quantum.Kata.Measurements {
     }
 
 
+    // Alternate reference implementation for task 1.9
+    // Slightly more expensive, but uses built-in functions
+    operation SuperpositionMeasurement_Alternate (qs : Qubit[], bits1 : Bool[][], bits2 : Bool[][]) : Int {
+        // measure all qubits and, treating the result as a number, check membership in bit arrays
+        let measuredState = ResultArrayAsInt(MultiM(qs));
+        for (s in bits1) {
+            if (BoolArrayAsInt(s)==measuredState) {
```
tcNickolas:
```suggestion
            if (BoolArrayAsInt(s) == measuredState) {
```

Measurements/ReferenceImplementation.qs
```diff
@@ -191,6 +191,20 @@ namespace Quantum.Kata.Measurements {
     }
 
 
+    // Alternate reference implementation for task 1.9
+    // Slightly more expensive, but uses built-in functions
+    operation SuperpositionMeasurement_Alternate (qs : Qubit[], bits1 : Bool[][], bits2 : Bool[][]) : Int {
+        // measure all qubits and, treating the result as a number, check membership in bit arrays
```
tcNickolas:
```suggestion
        // measure all qubits and, treating the result as an integer, check whether it can be found in one of the bit arrays
```

wsgac:
Done. Applied your suggestions.

PR #193, opened 2019-10-08
Add alternate reference implementation to Measurements task 1.10

wsgac:
I wrote an alternate reference implementation for Measurements problem 1.10. Using the built-in function `ResultArrayAsInt` I'm able to determine if the qubit register was in the `|0...0⟩` based on whether I get a zero.

### DIFF
--- a/Measurements/ReferenceImplementation.qs
+++ b/Measurements/ReferenceImplementation.qs
@@ -216,6 +216,13 @@ namespace Quantum.Kata.Measurements {
     }
 
 
+    // Alternate reference implementation for task 1.10
+    operation AllZerosOrWState_Alternate (qs : Qubit[]) : Int {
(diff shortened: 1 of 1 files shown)


Measurements/ReferenceImplementation.qs
```diff
@@ -216,6 +216,13 @@ namespace Quantum.Kata.Measurements {
     }
 
 
+    // Alternate reference implementation for task 1.10
+    operation AllZerosOrWState_Alternate (qs : Qubit[]) : Int {
+        // if we get 0 then the state is |0...0⟩
```
tcNickolas:
```suggestion
        // measure all qubits and convert the result into an integer;
        // if we get 0, the state was |0...0⟩, any non-0 integer indicates W state
```

wsgac:
Done. Thanks.

PR #194, opened 2019-10-08
Add alternate reference implementation to Measurements task 1.11

wsgac:
I'm using the observation that any measurement of GHZ state will either yield `0` or `2^n-1`.


wsgac:
Added the comment. Thanks.

PR #200, opened 2019-10-22
Added new tests to 1.8 and 1.9

Shaoqii:
Added new tests to 1.8 and 1.9, those tests have Length(bits1) > Length(bits2) tests

### DIFF
--- a/Measurements/Tests.qs
+++ b/Measurements/Tests.qs
@@ -349,12 +349,15 @@ namespace Quantum.Kata.Measurements {
 
         CheckSuperpositionBitstringsOneMeasurement(5, [30,14,10,6],     // [11110,01110,01010,00110]
                                                       [1,17,21,25]);    // [00001,10001,10101,11001]
-        CheckSuperpositionBitstringsMeasurement(2, [1,2],       // [01,10]
-                                                   [1]); // [01]
 
-        CheckSuperpositionBitstringsMeasurement(3, [7,5],       // [111,101]
-                                                   [2]); // [010]
+        CheckSuperpositionBitstringsOneMeasurement(2, [1,2],       // [01,10]
+                                                      [3]); // [11]
 
+        CheckSuperpositionBitstringsOneMeasurement(3, [5,7],       // [101,111]
+                                                      [2]); // [010]
+
+        CheckSuperpositionBitstringsOneMeasurement(4, [13,11,7,3], // [1101,1011,0111,0011]
+                                                      [2,4]); // [0010,0100]
     }
 
     // ------------------------------------------------------
@@ -386,14 +389,13 @@ namespace Quantum.Kata.Measurements {
                                                    [1,17,21,27]); // [00001,10001,10101,11011]
 
         CheckSuperpositionBitstringsMeasurement(2, [2,1],       // [10,01]
-                                                   [1]); // [01]
-
-        CheckSuperpositionBitstringsMeasurement(2, [1,2],       // [01,10]
-                                                   [1]); // [01]
+                                                   [3]); // [11]
 
         CheckSuperpositionBitstringsMeasurement(3, [7,5],       // [111,101]
                                                    [2]); // [010]
-
+                                                   
+        CheckSuperpositionBitstringsMeasurement(4, [13,11,7,3], // [1101,1011,0111,0011]
+                                                   [5,2]); // [0101,0010]
     }
 
 
--- a/Measurements/Tests.qs
+++ b/Measurements/Tests.qs
@@ -349,6 +349,17 @@ namespace Quantum.Kata.Measurements {
 
         CheckSuperpositionBitstringsOneMeasurement(5, [30,14,10,6],     // [11110,01110,01010,00110]
                                                       [1,17,21,25]);    // [00001,10001,10101,11001]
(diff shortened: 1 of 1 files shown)


Shaoqii:
I just made some Length(bits1) > Length(bits2) test cases for 1.8 and 1.9!

Measurements/Tests.qs
```diff
@@ -378,6 +389,22 @@ namespace Quantum.Kata.Measurements {
...
+        CheckSuperpositionBitstringsMeasurement(2, [1,2],       // [01,10]
+                                                   [1]); // [01]
+
+        CheckSuperpositionBitstringsMeasurement(3, [7,5],       // [111,101]
+                                                   [2]); // [010]
+
+        CheckSuperpositionBitstringsMeasurement(4, [13,11,7,3], // [1101,1011,0111,0011]
+                                                   [5,2]); // [0101,0010]
+
+        CheckSuperpositionBitstringsMeasurement(5, [14,17,9], // [01110,10001,01001]
```
tcNickolas: The testing harness doesn't support preparation of superposition states with 3 terms in the superposition (you can see this in `StatePrep_BitstringSuperposition` operation which does nothing if Length(bits) is anything but 1, 2 or 4). This is a good point - we need to improve the test harness either to cover other lenths or to fail early instead of providing the user invalid inputs. 

Long-term we might want to look into using [StatePreparationPositiveCoefficients](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.preparation.statepreparationpositivecoefficients) library operation to prepare the superposition states instead of writing the cases by hand.

For now, could you please delete this test and add the following line to the beginning of `StatePrep_BitstringSuperposition` operation (line 256, right after variable L is defined):

    Fact(L == 1 or L == 2 or L == 4, "State preparation only supports arrays of 1, 2 or 4 bit strings");

Measurements/Tests.qs
```diff
@@ -378,6 +389,22 @@ namespace Quantum.Kata.Measurements {
 
         CheckSuperpositionBitstringsMeasurement(5, [30,14,10,7],  // [11110,01110,01010,00111]
                                                    [1,17,21,27]); // [00001,10001,10101,11011]
+
+        CheckSuperpositionBitstringsMeasurement(2, [2,1],       // [10,01]
```
tcNickolas: Please make sure that the input vectors match the constraints described in `Tasks.qs`. For this task, `The only constraint on the bit strings is that all bit strings in the two arrays are distinct. `, so all numbers in the arrays have to be distinct, you can't have element 1 in both arrays.

Measurements/Tests.qs
```diff
@@ -349,6 +349,17 @@ namespace Quantum.Kata.Measurements {
 
         CheckSuperpositionBitstringsOneMeasurement(5, [30,14,10,6],     // [11110,01110,01010,00110]
                                                       [1,17,21,25]);    // [00001,10001,10101,11001]
+        CheckSuperpositionBitstringsMeasurement(2, [1,2],       // [01,10]
```
tcNickolas: Please note that this task should use `CheckSuperpositionBitstringsOneMeasurement` operation for testing each individual test case instead of `CheckSuperpositionBitstringsMeasurement` used by the next task.

Measurements/Tests.qs
```diff
@@ -349,6 +349,17 @@ namespace Quantum.Kata.Measurements {
...
+                                                   [1]); // [01]
+
+        CheckSuperpositionBitstringsMeasurement(3, [7,5],       // [111,101]
+                                                   [2]); // [010]
+
+        CheckSuperpositionBitstringsMeasurement(4, [13,11,7,3], // [1101,1011,0111,0011]
```
tcNickolas: I think this test violates the constraint on the input for the task?

    // You are guaranteed that there exists an index of a qubit Q for which 
    //  - all the bit strings in the first array have the same value in this position (all bits1[j][Q] are the same),
    //  - all the bit strings in the second array have the same value in this position (all bits2[j][Q] are the same),
    //  - these values are different for the first and the second arrays.

Shaoqii: done

Measurements/Tests.qs
Shaoqii: done

Measurements/Tests.qs
Shaoqii: done

Measurements/Tests.qs
Shaoqii: done

Measurements/Tests.qs
```diff
@@ -349,6 +349,15 @@ namespace Quantum.Kata.Measurements {
 
         CheckSuperpositionBitstringsOneMeasurement(5, [30,14,10,6],     // [11110,01110,01010,00110]
                                                       [1,17,21,25]);    // [00001,10001,10101,11001]
+
+        CheckSuperpositionBitstringsOneMeasurement(2, [1,2],       // [01,10]
```
tcNickolas: This test still violates the constraint on the input for task 1.8 - bit strings 01 and 10 don't have a position in which they both are 0 or both are 1.
```suggestion
        CheckSuperpositionBitstringsOneMeasurement(2, [0,2],       // [00,10]
```

PR #202, opened 2019-10-24
Updating to .NET Core 3.0 and newest Q# packages

bettinaheim:


bettinaheim:
> .ipynb files, scripts/install-iqsharp.ps1. I believe Andres has a script to do this change as automatically as possible.

Well, we have the problem that we don't have the right packages yet - we need the Katas updated for the signed build, and we need the signed build to get the packages to point to here. We are working on resolving that circular dependency now.

tcNickolas:
Closing this PR, since it is superseded by https://github.com/microsoft/QuantumKatas/pull/218.

PR #207, opened 2019-10-27
Updated reference sheet

sdf94:
I will adjust the tiles/fontawesome when the info is fixed.


quickref/qsharp-quick-reference.tex
```diff
@@ -165,22 +175,25 @@
     Call adjoint \newline operation & \texttt{Adjoint \emph{Name}(\emph{parameters})} \\
     Call controlled \newline operation & \texttt{Controlled \emph{Name}(\emph{controlQubits}, \emph{parameters})} \\
   \end{keysref}
-
   \begin{keysref}{Symbols and Variables}
     Declare immutable \newline symbol & \texttt{let \emph{name} = \emph{value}} \\
     Declare mutable \newline symbol (variable)  & \texttt{mutable \emph{name} = \emph{initialValue}} \\
     Update mutable \newline symbol (variable)   & \texttt{set \emph{name} = \emph{newValue}} \\
+    Apply-and-Reassign & \texttt{mutable \emph{name} = \emph{initialValue}} \newline
```
tcNickolas: This entry needs a general case before we can go into the example (something like "set *name* *operator*= *expression*"), and I don't think we should include a loop in the example - this might be confusing, since one can use apply-and-reassign outside of the loops.

quickref/qsharp-quick-reference.tex
```diff
@@ -165,22 +175,25 @@
...
+    \texttt{for (\emph{index} in \emph{range}) \{ } \newline
+    \texttt{ set counter += \emph{index};} \texttt{\}} \\ 
   \end{keysref}
 
-  \columnbreak%\
-
   \begin{keysref}{Arrays}
```
tcNickolas: Arrays section needs syntax for array element assignment, per https://docs.microsoft.com/en-us/quantum/language/statements#update-and-reassign-statement

quickref/qsharp-quick-reference.tex
```diff
@@ -165,22 +175,25 @@
...
     Allocation           & \texttt{mutable \emph{name} = new \emph{Type}[\emph{length}]} \\
     Length               & \texttt{Length(\emph{name})} \\
     k-th element         & \texttt{\emph{name}[k]} \newline NB: indices are 0-based  \\
     Array literal        & \texttt{[\emph{value0}, \emph{value1}, ...]} \newline
                            e.g.: \texttt{[true, false, true]} \\
-    Slicing (subarray)   & \texttt{\emph{name}[\emph{start}..\emph{end}]} \\
+    Slicing (subarray)   & \texttt{\emph{name}[\emph{start}...\emph{end}]} \newline
```
tcNickolas: This section gives the syntax but doesn't clarify what different types of syntax do - I think it would be more helpful if the left column had an example of an array we're slicing, and the right column listed both syntax used and the result you'll get. You can also highlight slicing with steps other than 1 this way very nicely.

quickref/qsharp-quick-reference.tex
```diff
@@ -198,15 +211,22 @@
     Return a value     & \texttt{return \emph{value}} \\
     Stop with an error & \texttt{fail "\emph{Error message}"} \\
   \end{keysref}
-
+   
+   \begin{keysref}{Conjugations}
+	ApplyWith & \texttt{operation \emph{Name}(\emph{in0} : \emph{type0}, ...) : \emph{returnType} \{} \newline
```
tcNickolas: I would call this entry `apply ... within`, since `ApplyWith` is a name of the function we used to have for this functionality, not functionality itself, and limit it to the actual `within ... apply` entries.

I would also provide a bit more context on what it does (i.e., that it is equivalent to withinBlock - applyBlock - Adjoint withinBlock sequence)

quickref/qsharp-quick-reference.tex
```diff
@@ -198,15 +211,22 @@
...
   \begin{keysref}{Debugging}
     Print a string     & \texttt{Message("Hello Quantum!")} \\
     Print an \newline interpolated string   
                        & \texttt{Message(\$"Value = \{\emph{val}\}")} \\
     Assert that a qubit is in $\ket{0}$ or $\ket{1}$ state   & \texttt{AssertQubit(Zero, \emph{oneQubit})}\\
     Print amplitudes \newline of wave function & \texttt{DumpMachine("dump.txt")} \\
   \end{keysref}
-    
+  
   \begin{keysref}{Qubit Allocation}
```
tcNickolas: Let's add syntax for allocating a mix of qubits and qubit registers to this section, since it's a very convenient shorthand.

sdf94: Could you clarify this?

quickref/qsharp-quick-reference.tex
```diff
@@ -139,14 +139,24 @@
   \end{keysref}
 
   \begin{keysref}{Derived Types}
-    Arrays                  & \texttt{\emph{elementType}[]} \\
+    Arrays                  & \texttt{\emph{elementType}[]} \newline
+    							\texttt{(elementType,elementType)[][] } \\
```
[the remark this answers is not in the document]
sdf94: I did not know that all spaces show the same way in all editors!

quickref/qsharp-quick-reference.tex
```diff
@@ -139,14 +139,24 @@
...
     Tuples                  & \texttt{(\emph{type0}, \emph{type1}, ...)} \newline
                               e.g.: \texttt{(Int, Qubit)} \\
     Functions               & \texttt{\emph{input} -> \emph{output}} \newline
                               e.g.: \texttt{ArcCos : (Double) -> Double} \\
     Operations              & \texttt{\emph{input} => \emph{output} : \emph{variants}} \newline
-                              e.g.: \texttt{H : (Qubit => Unit : \newline Adjoint, Controlled)} \\
+                              e.g.: \texttt{H : (Qubit => Unit : \newline Adj, Ctl)} \\
```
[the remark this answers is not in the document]
sdf94: Is this how it always will be? Also the new shorthand does not apply to the operators right? E.g. you can use Adjoint(q); or Controlled(q); , and these will not change to Adj(q); or Ctl(q);.

quickref/qsharp-quick-reference.tex
```diff
@@ -139,14 +139,24 @@
...
+  \begin{keysref}{User-Defined Types}
+  newtype  & \texttt{newtype Name = (Type, Type);} \newline
+            \texttt{e.g. newtype Complex = (Double, Double);} \\
+  nested newtype & \texttt{newtype Name1 = (type, (Name2 : type, type)); } \newline
+            \texttt{e.g. newtype Nested = (Double, (ItemName : Int, String));} \\
+  
+  \end{keysref}
+  
 
   \begin{keysref}{Functions, Operations and Types}
```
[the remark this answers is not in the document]
sdf94: I am not sure what you mean by remove types from this section. 

For the 2nd question, do you mean something like this where it uses the auto keyword?
`operation PrepareEntangledPair(here : Qubit, there : Qubit) : Unit {
    body (...) { // default body specialization
        H(here);
        CNOT(here, there);
    }

    adjoint auto; // auto-generate adjoint specialization
    controlled auto; // auto-generate controlled specialization
    controlled adjoint auto; // auto-generate controlled adjoint specialization
}`

quickref/qsharp-quick-reference.tex
tcNickolas: Something like `using (x, y) = (Qubit[N], Qubit())` to show that you can allocate multiple variables in one `using` statement

quickref/qsharp-quick-reference.tex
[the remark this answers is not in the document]
tcNickolas: This section says "Functions, Operations and Types" and has an entry for defining a UDT. Now that we're creating a new section devoted solely to UDTs, we should call this section "Functions and Operations" and remove the entry for defining a UDT from it, since it would be duplicate

For the second comment, if you want to define adjoint and controlled variants for the operation manually, you have to spell out the whole `adjoint {...}` syntax; but if you want to define them automatically based on the body, you can say `adjoint auto` (as you correctly pointed out), or you can just say `operation PrepareEntangledPair(here : Qubit, there : Qubit) : Unit is Adj+Ctl`, and the compiler will figure out that it's a shorthand for `adjoint auto; controlled auto;`. This is a very convenient syntax, since it allows to skip `body(...)` part and save one level of nesting in the code.

quickref/qsharp-quick-reference.tex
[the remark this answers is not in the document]
tcNickolas: They do (well, some editors highlight them but at least they always occupy a single space). Tabs show differently - GitHub shows them as taking 8 spaces, and other editors typically use 2 or 4 spaces, so it can be less noticeable in local editor but very noticeable in a pull request on GitHub.

quickref/qsharp-quick-reference.tex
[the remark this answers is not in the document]
tcNickolas: Yes, unless we come up with a different syntax (Q# as a language is still in preview, so breaking changes are possible). Yes, this is just for operation signatures, to specify that the operation is, has to be adjointable/controllable. When you call a variant of the operation, it's still `Adjoint q` or `Controlled q`.

sdf94:
The font is not considered because it wouldn't recognize the fontawesome page. I think it is more important to figure out the changes needed to make in the text before I attempt to figure out why it isn't working. :)

Issue #187, opened 2019-10-06
Improve testing harness in RippleCarryAdder kata, task 1.7


tcNickolas:
Most tasks in this kata take inputs in an arbitrary superposition state and require implementing the right unitary transformation on them; to verify the solutions, the tests compare the learner's solution to the reference one as unitaries using `AssertOperationsEqualReferenced`. (Sometimes one or several inputs are in |0⟩ state, but this does not affect the expected solution, so we still can use the assert.)

In task 1.7, however, one of the inputs is guaranteed to be in |0...0⟩ state, and the "challenge" solution relies on that fact to reuse these qubits instead of allocating new ones. This means that the main solution and the reference ones perform different transformations, even though they are the same on the basis vectors. The test harness for this task uses only `AssertOperationImplementsBinaryFunction` to verify the solution instead of both it and `AssertOperationsEqualReferenced`.

We need to figure out a way to add verification of the solution as a unitary in this task. I think it should be possible to implement a wrapper operation which is a unitary on inputs a and b and allocates qubits in |0...0⟩ state for target qubits before calling the adder. Both solutions, when wrapped in such an adder, should implement the same unitary, and can be compared using `AssertOperationsEqualReferenced`.

* We can probably do the same for other tasks of this kata in which some of the inputs are in |0⟩ state, but it makes sense to start with this one. Afterwards we can either modify the testing harnesses for them similarly or modify the task descriptions to remove the promise of |0⟩ state in inputs.

sterilli-ms:
I'm looking at this one

PR #213, opened 2019-10-29
Katas issue #187

sterilli-ms:
Added new wrapper to allocate Qubits to 0 state.  Would love feedback.


RippleCarryAdder/Tests.qs
```diff
@@ -235,16 +242,16 @@ namespace Quantum.Kata.RippleCarryAdder {
...
         for (i in 1 .. 4){
-            let testOp = QubitArrayAdderWrapper(i, ArbitraryAdder, _);
-            AssertOperationImplementsBinaryFunction(testOp, BinaryAdder(_, i), 2 * i, i + 1);
+			let testOp1 = QubitArrayAdderWrapper(i, ArbitraryAdder, _);
+
+            let testOp2 = QubitArrayAllocateAdderWrapper(i, ArbitraryAdder, _);
+			let refOp = QubitArrayAllocateAdderWrapper(i, ArbitraryAdder_Reference, _);
+
+			AssertOperationImplementsBinaryFunction(testOp1, BinaryAdder(_, i), 2 * i, i + 1);
+            AssertOperationsEqualReferenced(i, testOp2, refOp);
```
tcNickolas: The way this check is written now doesn't actually test the solution, though it took me a while to figure out why exactly.

`AssertOperationsEqualReferenced` compares two operations assuming that they are operations on `i` qubits: it prepares `i` qubits in a certain entangled state, passes them to the first operation and then to the adjoint of the second operation and then checks that these qubits ended in the same state as they started.  `QubitArrayAllocateAdderWrapper` wraps the operations we want to compare, but it doesn't actually use the qubits passed to them as the last parameter, it allocates fresh qubits for all parameters and runs the operations on them. This means that it has no effect on the qubits passed to it, and any operation will be equal to any other operation. Indeed, if you comment out `AssertOperationImplementsBinaryFunction` line above, you'll see that any operation, including empty one, will pass the test!

---

I did some experimenting with the code to see how to get the behavior we need, and it is a bit trickier than I expected when I was writing up the issue, so let me go into more details here.

We need to compare two operations which take parameters `a, b, sum, carry` each; of these parameters `a, b, carry` are allowed to be in arbitrary state, and `sum` is guaranteed to be in 0 state. We can do this using the following modification of the [`AssertOperationsEqualReferenced` implementation](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Intrinsic/Diagnostics/AssertOperationsEqualReferenced.qs):

    using ((reference, target, sum) = (Qubit[2*i+1], Qubit[2*i+1], Qubit[i])) {
        let a = target[0..i-1];
        let b = target[i..2*i-1];
        let carry = Tail(target);

        _prepareEntangledState(reference, target);
        ArbitraryAdder(a, b, sum, carry);
        Adjoint ArbitraryAdder_Reference(a, b, sum, carry);
        Adjoint _prepareEntangledState(reference, target);
        AssertAllZero(reference + target);
    }

You can see that the parameters that accept arbitrary states are passed to `_prepareEntangledState` following the same verification logic as is used for unitary comparison, and the parameter that has to start in 0 state is left alone with only the operations applied to it.

tcNickolas:
Will you have time to work on this PR in the next two weeks? I'd really like to complete it by the end of the year, just to keep things nice and organized :-)

PR #215, opened 2019-10-30
Tasks 6 and 7 uploaded to Superposition Workbook

ulitoo:


tcNickolas:
I brushed up presentation for task 6 a bit, will continue on task 7 tomorrow. One comment about the image `Task6OutputHadamardasControl.png` - could you please update it to say "flip" instead of "swap"? Swap is a two-qubit gate, and I'm worried that using it will confuse the reader instead of enlightening them...

Tagging @jimcristofono who authored the picture.

PR #217, opened 2019-11-01
Superposition workbook (fix for task 2)

jimcristofono:
Apparent case sensitivity problem.  The notebook references the image file in lower case  .  Renamed the file so that the file name matches the reference in the notebook.


jimcristofono:
Messed up the pull request.  Will recreate it.

PR #218, opened 2019-11-01
Updating to latest release (0.10.1910.3107)

anpaz-msft:


anpaz-msft:
Binder checks: https://mybinder.org/v2/gh/Microsoft/QuantumKatas/release/v0.10

NuGet.config
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tcNickolas: It doesn't seem like we should need this file for accessing the published NuGets?

tutorials/SingleQubitGates/SingleQubitGates.csproj
```diff
@@ -6,9 +6,9 @@
   </PropertyGroup>
```
tcNickolas: Should line 3 here be updated to `netcoreapp3.0` as well? This seems to fail CI (Sorry, GitHub doesn't let me leave a comment there)

NuGet.config
anpaz-msft: Not for the released, but yes for the pre-released. We are adding a similar file to all our repos so it's easier to integrate x-repo changes before they become part of a release.

tutorials/SingleQubitGates/SingleQubitGates.csproj
anpaz-msft: yeap. this project was added after the migration of the rest was done so it was not picked up.
I'll update.

Issue #204, opened 2019-10-24
Write PowerShell script to update QDK to the new version


tcNickolas:
Updating QDK to the next version requires doing a fixed set of changes to an ever-growing set of files (see recent #179 and the most recent #235 for a list of files and changes). 

Currently we're using a generic Bash script to automate most of the change (see attached [updateQDKVersion.txt](https://github.com/microsoft/QuantumKatas/files/3769677/updateQDKVersion.txt)), but it would be nice to have a PowerShell script tailored to the Katas project.

sdf94:
I will take this!

tcNickolas:
Great, thank you!

sdf94:
Could you give me a zipped up version of packages.config and the other folder for testing?

tcNickolas:
Sorry, I'm not sure what exactly you need? To test the script, you can just run it on the master branch of the Katas with some arbitrary version like 1.2.3.4 and check that the version number has been updated in all the files covered in #235 (except .code-workspace file which was deleted independently), you shouldn't need anything extra.

PR #244, opened 2019-11-23
Change QDK Version powershell script

sdf94:
This pull request is in response to issue #204 .


sdf94:
I am working on powershell on VSCode on a mac and have a difficulty re-creating the error that you got (where it changes every little thing to the version). However, I have updated the script to include the individual versions of each katas.

updateQDKVersion.ps1
```diff
@@ -0,0 +1,45 @@
...
+$csproj_string = 'Include="{0}" Version=' -f $pkgs[$i]
+Write-Host $csproj_string
+$file = (Get-ChildItem -recurse -Name -include "BasicGates/*.csproj")
+Write-Host $file
+$sentence = (Get-ChildItem -recurse | Select-String -pattern $csproj_string -Context 0).Line
+Write-Host (($sentence -split 'Version=')[1] | %{$_.split('"')[1]})
+$csproj_old = (($sentence -split 'Version=')[1] | %{$_.split('"')[1]})
+Write-Host  "Changing Version from $csproj_old to $ver"
+
+for .csproj files - <PackageReference Include="Microsoft.Quantum.???" Version="___" />
```
tcNickolas: PowerShell  doesn't recognize the lines starting with "for", which makes sense - they look like comments and not the actual PowerShell commands? My comment in the previous review was to change the logic of detecting the versions of the packages that need to be updated, from finding the old version in one project and replacing it across all other projects to finding all old versions in each of the contexts we have and replacing them with the new version, regardless of what the old version was.

updateQDKVersion.ps1
```diff
@@ -0,0 +1,45 @@
...
+for .ipynb files - "%package Microsoft.Quantum.???::___"
+for Dockerfile - FROM mcr.microsoft.com/quantum/iqsharp-base:___
+for install-iqsharp.ps1 - dotnet tool install Microsoft.Quantum.IQSharp --version ___ --tool-path $Env:TOOLS_DIR
+
+$configFiles = (Get-ChildItem -recurse -include "*.csproj", "*.ipynb", "install-iqsharp.ps1","Dockerfile")
+foreach ($file in $configFiles)
+{
+    (Get-Content $file.PSPath) |
```
tcNickolas: This sequence of commands to replace the version is interesting... The error I got last time, with new version inserted between each pair of characters, does not happen again (which is nice but still worries me a bit, since I don't understand where it came from and why it went away). Instead, all `.csproj` files switched their encoding from "UTF-8 with signature" to "Western European (Windows)". 

I remember I struggled with preserving the encoding of the updated files in [this PowerShell script](https://github.com/microsoft/QuantumKatas/blob/master/scripts/fix-unicode.ps1), but I tried the same approach here and it didn't work... Any ideas?

sdf94:
Can you verify what the encoding actually is?

updateQDKVersion.ps1
sdf94: The $_ was creating the duplication.

tcNickolas:
Looks like .csproj files have UTF8 encoding, and the rest of them have a different one (Western European?). My PowerShell doesn't know encoding "windows-1251".

sdf94:
Okay great. This is progress. I think it might be because my computer is in
NL. Would utf8 coding work? If yes, I will force it to that encoding.

tcNickolas:
I'm not sure how well the Jupyter Notebooks will take to being forced to UTF8... I would try to just do different things for different files - read and write in UTF8 for .csproj files, and read and write in default encoding for the rest of them.

I finally had some time today to try and figure out the mystery of the new version inserted between all pairs of characters in strings, and other things that puzzle me about PowerShell. Here is a snippet I came up with for dealing with .csproj files that goes through each file just once (your current code seems to go through each file once per every possible Microsoft.Quantum package, which takes longer than I'd like an update script to work) and preserves the encoding:

```powershell
$files = Select-String -Path "**\*.csproj" -pattern "Microsoft.Quantum" | Select-Object -Unique Path

$files | ForEach-Object {
    (Get-Content -Encoding UTF8 $_.Path) | ForEach-Object {
         $isQuantumPackage = $_ -match 'PackageReference Include=\"Microsoft\.Quantum\.[a-zA-Z\.]+\" Version=\"(?<oldVersion>[0-9|\.]+)\"'
         if ($isQuantumPackage) {
             $_ -replace $Matches.oldVersion, '1.2.3.4'
         } else {
             $_
         }
}
```

It reads the file using UTF8 encoding and checks each string for being a package reference for one of Microsoft.Quantum packages (not really caring which one it is); if the string is a package reference, it replaces old package version with the new one, if it's not it keeps the string as is (and I think *that* was the source of that mystifying error - trying to replace old version with new if old version was not found in a string).

What do you think about this snippet? I think we could use it for .csproj (once we parametrize the new version - right now it is hardcoded), and use a similar approach for the rest of the files, removing UTF8 switch and using the right match string for each type of files.

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
...
+             $_
+         }
+    } | Set-Content $_.Path
+}
+
+$ps1_string = 'Microsoft.Quantum.IQSharp --version (?<oldVersion>[0-9|\.]+)'
+$ps1Files = (Select-String -Path "**\install-iqsharp.ps1" -pattern "Microsoft.Quantum.IQSharp" | Select-Object -Unique Path)
+$ps1Files | ForEach-Object {
+    
+    (Get-Content -Encoding 'Windows-1252' $_.Path) | ForEach-Object {
```
tcNickolas: My PowerShell doesn't recognize encoding Windows-1252:

    Cannot bind parameter 'Encoding'. Cannot convert value "Windows-1252" to type
    "Microsoft.PowerShell.Commands.FileSystemCmdletProviderEncoding". Error: "Unable to match the identifier name
    Windows-1252 to a valid enumerator name. Specify one of the following enumerator names and try again:
    Unknown, String, Unicode, Byte, BigEndianUnicode, UTF8, UTF7, UTF32, Ascii, Default, Oem, BigEndianUTF32"

I tried just omitting the encoding here and it worked just fine. Do you get any problems on your system if you omit the encoding?

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
+$ver = $args[0]
+Write-Host $ver
+
+$csString = 'PackageReference Include=\"Microsoft\.Quantum\.[a-zA-Z\.]+\" Version=\"(?<oldVersion>[0-9|\.]+)\"'
+$csFiles = (Select-String -Path "**\*.csproj" -pattern "Microsoft.Quantum" | Select-Object -Unique Path)
```
tcNickolas: I think this mask for the Path doesn't catch the files in `tutorials` subfolder, since they are nested one level deeper than the others. Do you know what PowerShell mask corresponds to "a file arbitrarily deep in the tree"? (Same change will need to be applied to the `.ipynb` pattern)

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
...
+             $_
+         }
+    } | Set-Content -Encoding UTF8 $_.Path
+}
+
+$ipynb_string = '%package Microsoft.Quantum.Katas::(?<oldVersion>[0-9|\.]+)'
+$ipynbFiles =  (Select-String -Path "**\*.ipynb" -pattern "Microsoft.Quantum" | Select-Object -Unique Path)
+
+$ipynbFiles | ForEach-Object {
+    if ($_)
```
tcNickolas: Why do we need this check? I thought none of the results we loop over should be empty?..

tcNickolas:
@cgranade You mentioned you were willing to take a look at this pull request from the PowerShell point of view - I think we're getting close functionality-wise, so now would be a good time, if you're still willing :-)

cgranade:
> @cgranade You mentioned you were willing to take a look at this pull request from the PowerShell point of view - I think we're getting close functionality-wise, so now would be a good time, if you're still willing :-)

Absolutely, one minute and I'll get that review in. Thanks for flagging me!

updateQDKVersion.ps1
cgranade: One way to explicitly get arbitrarily deep files would be with `Get-ChildItem -Recurse`:
```powershell
Get-ChildItem -Recurse *.csproj `
    | ForEach-Object { Select-String -Path $_ -Pattern "Microsoft.Quantum" } `
    | Select-Object -Unique Path
```

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
...
+$csFiles = (Select-String -Path "**\*.csproj" -pattern "Microsoft.Quantum" | Select-Object -Unique Path)
+$csFiles | ForEach-Object {
+    (Get-Content -Encoding UTF8 $_.Path) | ForEach-Object {
+         $isQuantumPackage = $_ -match $csString
+         if ($isQuantumPackage) {
+             $_ -replace $Matches.oldVersion, $ver
+         } else {
+             $_
+         }
+    } | Set-Content -Encoding UTF8 $_.Path
```
cgranade: For compatibility reasons, this line writes out using an old nonstandard variant of UTF8 that has a byte-order mark (BOM) at the beginning. On PowerShell 6 and later, `-Encoding UTF8NoBom` can be used to suppress the BOM. We've before had cases where some tools fail to recognize UTF8-encoded files with the BOM, unfortunately.

At this point, PowerShell 5.1 and earlier (also known as Windows PowerShell) are fairly old, such that I think we can use `UTF8NoBom` fairly safely. To be sure, you can include a `#Requires -PSEdition Core` statement at the top of the script to make sure that the `UTF8NoBom` value is supported.

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
...
+}
+
+$ipynb_string = '%package Microsoft.Quantum.Katas::(?<oldVersion>[0-9|\.]+)'
```
cgranade: Minor, but PowerShell style tends to favor `$camelCase` variable names to `$snake_case`, such as in your use of `$isQuantumPackage` above.
```suggestion
$ipynbString = '%package Microsoft.Quantum.Katas::(?<oldVersion>[0-9|\.]+)'
```

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
...
+    if ($_)
+    {
+        (Get-Content $_.Path) | ForEach-Object {
+            $isQuantumPackage = $_ -match $ipynb_string
+            if ($isQuantumPackage) {
+                $_ -replace $Matches.oldVersion, $ver
+            } else {
+                $_
+            }
+        } | Set-Content $_.Path
```
cgranade: Similarly here, I'd suggest explicitly denoting the encoding.

updateQDKVersion.ps1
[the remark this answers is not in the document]
cgranade: I agree, no glob is needed here. I'd suggest using `Join-Path` and `$PSScriptRoot` together to explicitly refer to where the targeted file is in the repo.

cgranade: I'm with @tcNickolas on this; we should likely avoid Windows-1252 if possible, especially to help compatibility with non-Windows users.

updateQDKVersion.ps1
```diff
@@ -0,0 +1,61 @@
+$ver = $args[0]
```
cgranade: I'd suggest a shebang line here to help non-Windows users.
```suggestion
#!/usr/bin/env pwsh
$ver = $args[0]
```

cgranade: Generally, it's more robust in PowerShell to use a `param()` block to declare parameters for a script than to unpack `$args` manually. That allows for typechecking, automatic help, and tab completion. You can even use [`[CmdletBinding()]`](https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_functions_cmdletbindingattribute?view=powershell-7) to make sure that parameters are mandatory, providing built-in error messages when they aren't present.
```suggestion
[CmdletBinding()]
param(
    [Parameter(Mandatory=$true)]
    [string] $Version
);
```

cgranade: Last suggestion for the top of the file, but it's often helpful to put a brief documentation comment at the top of the script so that it can be used together with integrated help tools such as `Get-Help`, intelliSense in VS Code and PowerShell ISE, or dotnet-interactive's Jupyter-based help commands.
```suggestion
<#
    .SYNOPSIS
        Updates the contents of this repo to use a given version of
        the Quantum Development Kit.
        
    .PARAMETER Version
        The version that this repo should be updated to. This version should be
        a valid NuGet package version as well as a valid tag for the
        mcr.microsoft.com/quantum/iqsharp-base Docker image.
        
    .EXAMPLE
    
        PS> ./Update-QDKVersion.ps1 -Version 0.10.2002.2610
#>
$ver = $args[0]
```

updateQDKVersion.ps1
sdf94: So if we do go the recurse way, there might be ipynb that do not have a match. I ran into that problem before, so this is just a catch-all.

updateQDKVersion.ps1
sdf94: Do you know what encoding the quantum operators are? I tried to change it to ascii or utf8 but it wouldn't translate those operators appropriately. @tcNickolas do you know?

updateQDKVersion.ps1
sdf94: Nope. What encoding would you suggest @cgranade?

sdf94: @tcNickolas just noting the whole quantum operators here again. I changed it to jupyter notebook's native encoding, but that changed the quantum operators.

tcNickolas: Let's omit the encoding parameter in the commands for the non-csproj files if omitting the encoding allows to preserve them unchanged.

updateQDKVersion.ps1
tcNickolas: Let's omit the encoding parameter in the commands for the non-csproj files if omitting the encoding allows to preserve them unchanged.

updateQDKVersion.ps1
tcNickolas: Sorry, I missed this comment in the first read. I don't think 5.1 is _that_ old - I set up my laptop less than a year ago and it has 5.1 - so I don't think we want to phase it out completely. I kept UTF8 here and removed requirement for Core edition.

PR #250, opened 2019-12-18
Update QDK to version 0.10.1912.0501

tcNickolas:


tcNickolas:
Putting this update on hold while investigating #251.

tcNickolas:
I don't think we'll do a hotfix release before the next QDK release, so this pull request will not be needed. I'll close it but keep the branch alive for some time.

PR #253, opened 2019-12-21
[Measurement]: Refined test results in DistinguishTwoStates_OneQubit

ulitoo:
added 2 strings parameter to signal on the error what specific state was miscategorized, how many times and how.

### DIFF
--- a/Measurements/Tests.qs
+++ b/Measurements/Tests.qs
@@ -24,9 +24,8 @@ namespace Quantum.Kata.Measurements {
     // with Bool return
     operation DistinguishTwoStates_OneQubit (statePrep : ((Qubit, Int) => Unit), testImpl : (Qubit => Bool), stateName : String[]) : Unit {
         let nTotal = 100;
-        let Nstate = Length(stateName);
-        mutable misclassifications = new Int[Nstate];
-        mutable Totalmisclassifications = 0;
+        let nStates = 2;
+        mutable misclassifications = new Int[nStates];
         
         using (qs = Qubit[1]) {
             for (i in 1 .. nTotal) {
@@ -41,7 +40,7 @@ namespace Quantum.Kata.Measurements {
                 // get the solution's answer and verify if NOT a match, then differentiate what kind of mismatch
                 let ans = testImpl(qs[0]);
                 if (ans != (state == 1)) {
-                    set misclassifications w/= state <-  misclassifications[state]+1;
+                    set misclassifications w/= state <- misclassifications[state] + 1;
                 }
 
                 // we're not checking the state of the qubit after the operation
@@ -49,15 +48,16 @@ namespace Quantum.Kata.Measurements {
             }
         }
         
-        for (i in 0 .. Nstate-1) {
+        mutable totalMisclassifications = 0;
+        for (i in 0 .. nStates - 1) {
             if (misclassifications[i] != 0) {
-                set Totalmisclassifications += misclassifications[i];
-                Message($"'Misclassified' {stateName[i]} as {stateName[(i+1)%Nstate]} in {misclassifications[i]} test runs.");   
+                set totalMisclassifications += misclassifications[i];
+                Message($"Misclassified {stateName[i]} as {stateName[1 - i]} in {misclassifications[i]} test runs.");   
             }
         }
         
-        // This Check will tell the # of Failed clasifications
-        EqualityFactI(Totalmisclassifications, 0, $"{Totalmisclassifications} test runs out of {nTotal} returned incorrect state (see output for details).");
+        // This check will tell the total number of failed classifications
+        Fact(totalMisclassifications == 0, $"{totalMisclassifications} test runs out of {nTotal} returned incorrect state (see output for details).");
     }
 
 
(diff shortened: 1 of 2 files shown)


Measurements/Tests.qs
```diff
@@ -22,9 +22,10 @@ namespace Quantum.Kata.Measurements {
 
     // "Framework" operation for testing single-qubit tasks for distinguishing states of one qubit
     // with Bool return
-    operation DistinguishTwoStates_OneQubit (statePrep : ((Qubit, Int) => Unit), testImpl : (Qubit => Bool)) : Unit {
+    operation DistinguishTwoStates_OneQubit (statePrep : ((Qubit, Int) => Unit), testImpl : (Qubit => Bool), str1 : String, str_aim : String) : Unit {
```
tcNickolas: I'd pass string representations as an array, indexed by the same integer as used to pass to `statePrep` - this way it will be harder to confuse which string corresponds to which representation.

Measurements/Tests.qs
```diff
@@ -22,9 +22,10 @@ namespace Quantum.Kata.Measurements {
...
         let nTotal = 100;
-        mutable nOk = 0;
+        mutable nNOTOk0as1 = 0;
```
tcNickolas: Similarly here, I'd make this an array `misclassifications` and store how many of states with the given index were misclassified. Then lines 43-48 become just `set misclassifications[state] += 1;`

Measurements/Tests.qs
```diff
@@ -36,18 +37,28 @@ namespace Quantum.Kata.Measurements {
...
             }
         }
 
-        EqualityFactI(nOk, nTotal, $"{nTotal - nOk} test runs out of {nTotal} returned incorrect state.");
+        // This Check will tell the # of Failed clasifications
+        // EqualityFactI((nNOTOk0as1 + nNOTOk1as0), 0, $"{nNOTOk0as1 + nNOTOk1as0} test runs out of {nTotal} returned incorrect state.");
+        
+        // This Check will give more granularity telling the # of misclasifications by type (0->1) vs (1->0)
+        // Gets two strings as input of what are the two states to be clasified, str_aim being the target state. 
+        EqualityFactI((nNOTOk0as1 + nNOTOk1as0), 0, $"'Misclassified' {str1} as {str_aim} in {nNOTOk0as1} test runs, and {str_aim} as {str1} in {nNOTOk1as0} test runs out of {nTotal} total runs.");
```
tcNickolas: I would split this check in two parts (having in mind that the later tasks will need to report errors on more states): 
* first go through all possible misclassifications (in this case iterate over the array) and report to output (using `Message`) only those that actually happened (for example, if the code always returns "|0⟩" state, we won't report "|0⟩ was misclassified as |1⟩ 0 times", we'll only point the actual mistakes done). At the same time we'd accumulate the total count of misclassifications during this iteration.
* after that do a `fail` if we have any misclassifications, saying something like "X test runs out of Y returned incorrect state (see output for details)".

Measurements/Tests.qs
```diff
@@ -119,14 +130,14 @@ namespace Quantum.Kata.Measurements {
         // cross-test
         // alpha = 0.0 or PI() => !isQubitOne
         // alpha = PI() / 2.0 => isQubitOne
-        DistinguishTwoStates_OneQubit(StatePrep_IsQubitOne, IsQubitA(PI() / 2.0, _));
+        DistinguishTwoStates_OneQubit(StatePrep_IsQubitOne, IsQubitA(PI() / 2.0, _),"|B⟩","|A⟩");
```
tcNickolas: For this task, since states A and B can be different in each test case, it would be helpful to give the exact representations of |A⟩ and |B⟩ rather than generic variable names. For example, if the learner has a solution that does rotation in different direction, they might still distinguish |0⟩ and |1⟩ correctly but get stuck on |+⟩ and |-⟩ or the states with unequal superposition.

Measurements/Tests.qs
ulitoo: Working on this one, thanks for the suggestion
Variable will be:     stateName : String[]

Measurements/Tests.qs
ulitoo: Thanks for the tip. Could not find a way to += in Q#, instead I will use:
                    set misclassifications w/= state <-  misclassifications[state]+1;

Measurements/Tests.qs
ulitoo: Makes sense. The detailed output will look like:

'Misclassified' |0⟩ as |1⟩ in 32 test runs out of 100 total runs.
'Misclassified' |1⟩ as |0⟩ in 22 test runs out of 100 total runs.
54 ≠ 0: 54 test runs out of 100 returned incorrect state (see output for details).
Try again!

Measurements/Tests.qs
ulitoo: Agreed. Output will look like this now:
'Misclassified' |B⟩=(-sin(1π/10)|0⟩ + cos(1π/10)|1⟩) as |A⟩=(cos(1π/10)|0⟩ + sin(1π/10)|1⟩) in 3 test runs.
'Misclassified' |A⟩=(cos(1π/10)|0⟩ + sin(1π/10)|1⟩) as |B⟩=(-sin(1π/10)|0⟩ + cos(1π/10)|1⟩) in 4 test runs.

ulitoo:
@tcNickolas , please have a look at the code now and let me know what you think.

Measurements/Tests.qs
tcNickolas: True, I keep forgetting the array element update syntax...

PR #290, opened 2020-03-01
Update QDK to version 0.10.2002.2610

tcNickolas:


cgranade:
I'd suggest also adding a call to the `.WithJupyterDisplay(this QuantumSimulator)` extension method in [`KataMagic.CreateSimulator`](https://github.com/microsoft/QuantumKatas/blob/mariia/update-qdk/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L141):

```csharp
using Microsoft.Quantum.IQSharp.Jupyter;

...

        /// <summary>
        /// Creates the instance of the simulator to use to run the Test 
        /// (for now always CounterSimulator from the same package).
        /// </summary>
        public virtual SimulatorBase CreateSimulator() =>
            new CounterSimulator().WithJupyterDisplay();
```

In the meantime, everything else looks good, currently trying out in Binder at https://mybinder.org/v2/gh/microsoft/quantumkatas/mariia/update-qdk.

cgranade:
I'm seeing some errors about missing namespaces in the build logs, is that typical?

```
[NbConvertApp] Converting notebook DeutschJozsaAlgorithm/DeutschJozsaAlgorithm.ipynb to markdown
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6104: No namespace with the name "Quantum.Kata.Utils" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "GetOracleCallsCount" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "GetOracleCallsCount" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "ResetOracleCallsCount" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "GetOracleCallsCount" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "ResetOracleCallsCount" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "ResetOracleCallsCount" exists.
```

cgranade:
I'm not sure if it's related, but I did find a bug in the `BellStateChange3` task of **BasicGates** when testing on Binder. Namely, the `Y` operation should map qubits initially in the |β₀₀⟩ state to the |β₁₁⟩ state, since |βᵢⱼ⟩ = 𝑋ⁱ𝑍ʲ |β₀₀⟩ is typically taken as the definition of |βᵢⱼ⟩, but is rejected as a solution to the task that asks users to prepare |β₁₁⟩. Since the task operation is declared `is Adj + Ctl`, I suspect the issue is that the task is checking controlled applications of the task operation, rather than the task operation itself, but that additional requirement isn't listed in the task description:

image

tcNickolas:
For the first comment, QDK version update should only include the version update and fixing the breaking changes introduced by it; I'm not sure what `WithJupyterDisplay` does (visual representation of DumpMachine outputs? I found the code at https://github.com/microsoft/iqsharp/blob/master/src/Jupyter/Extensions.cs but it doesn't have documentation), but it doesn't look like a it should be part of version update. I'm happy to do this change separately, once I understand what it does :-)

For the second comment, I think these are typical.

For the third comment, it's not a bug, it's a feature :-) Using Y gate introduces an extra global phase; I've found that new learners get very confused when the solution that introduces global phase is accepted, so all tasks in BasicGates require exact implementation, without extra global phase. I can make that requirement more pronounced.

cgranade:
> For the first comment, QDK version update should only include the version update and fixing the breaking changes introduced by it; I'm not sure what `WithJupyterDisplay` does (visual representation of DumpMachine outputs? I found the code at https://github.com/microsoft/iqsharp/blob/master/src/Jupyter/Extensions.cs but it doesn't have documentation), but it doesn't look like a it should be part of version update. I'm happy to do this change separately, once I understand what it does :-)

As per offline discussions, and comments on the original PR (https://github.com/microsoft/iqsharp/pull/81), that extension method is used to enable outputting rich displays such as that generated by DumpMachine and DumpRegister. That isn't currently part of the interface that target machines have with hosts, as per @anpaz-msft's suggestion that adding richer display output should be deferred to IQuantumProcessor. I take the point that it should be documented, though, and will make a PR to add that; it was made public late on in the PR to enable exactly this scenario as per your request, and I forgot to add the API documentation comment when doing moving it from private to public.

> For the second comment, I think these are typical.
> 
> For the third comment, it's not a bug, it's a feature :-) Using Y gate introduces an extra global phase; I've found that new learners get very confused when the solution that introduces global phase is accepted, so all tasks in BasicGates require exact implementation, without extra global phase. I can make that requirement more pronounced.

PR #291, opened 2020-03-02
Make iq# installation ore resilient

anpaz-msft:
QDK build has been failing when using the vs2019 build agent during the execution of this script.
This makes the installation of iq# more resilient to errors.

### DIFF
--- a/scripts/install-iqsharp.ps1
+++ b/scripts/install-iqsharp.ps1
@@ -18,13 +18,17 @@ try {
 }
 
 if ($install) {
-    Write-Host ("Installing Microsoft.Quantum.IQSharp at $Env:TOOLS_DIR")
-    dotnet tool install Microsoft.Quantum.IQSharp --version 0.10.1911.1607 --tool-path $Env:TOOLS_DIR
+    try {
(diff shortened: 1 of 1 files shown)


scripts/install-iqsharp.ps1
```diff
@@ -5,12 +5,27 @@ $ErrorActionPreference = 'Stop'
...
+
+Write-Host ("Installing IQ# tool.")
+$install = $False
+
+# Install iqsharp if not installed yet.
+try {
+    dotnet iqsharp --version
+    $install = ($LastExitCode -ne 0)
+} catch {
+    Write-Host ("`dotnet iqsharp --version` threw Exception.")
```
cgranade: Do we want to write out what the exception was in this case?

tcNickolas: I added logging the error message in this case similar to the other scenario.

tcNickolas:
Kicked the CI build manually at https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=8586&view=results; the build passed, so merging the PR.

PR #306, opened 2020-03-19
Mbu/superposition

mbuchberger1967:
Superposition/Workbook_Superposition_Part2.ipynb is the relevant change: 
Superposition Task12-Hardy State : added math explanation to Workbook


mbuchberger1967:
Hi,
1. I resolved the conflicts, and merged master into this branch. But unfortunately there are still other "changed files" due to the mess I had in my branches.
I think I will close this PR and do a new one, with the right files modified.

2. Regarding your content proposal.
Yes, the matrices and vectors are redundant somehow to the Dirac notation. I am not a mathematic guy nor a physic guy, I'm a software developer. For me the Dirac notation is not so easy to handle, still, though I'm using it now more often :)
This is one of the Katas from the beginning and I could imagine that other beginners have also problems to handle Dirac notations...
I needed exactly this matrice presentation to understand all the rotations and also the effect of a Zero-Controlled operation on the control qbit.

To pay respect to your proposals I'll try the following.
I will rework my description and try to do everything in Dirac notation, and have hints to expand that show the corresponding matrice operations. What is your opinion on that?

BR Markus

mbuchberger1967:
Hi Mariia,
I have another question regarding a different KATA and dont know how to handle that. As I had only contact with you so far, I write that comment here hoping you can reply on my point.

In the MagicSquareGame Kata, Task 2.1 "Create Entangled State" I guess there is an error:
I wanted to ask you if you agree, and how I shall handle that error correctly (shall I open an issue?)

Here are the details (I hope I dont bother you too much). Feel free to suggest me whom else to contact.

    // Task 2.1. Entangled state
    // Input: An array of 4 qubits in the |0000⟩ state.
    // Goal:
    //     Create the entangled state
    //     |ψ⟩ = ((|+⟩₀ ⊗ |+⟩₂ + |-⟩₀ ⊗ |-⟩₂) / sqrt(2)) ⊗ ((|+⟩₁ ⊗ |+⟩₃ + |-⟩₁ ⊗ |-⟩₃) / sqrt(2)),
    //     where |ψ⟩₀ and |ψ⟩₁ are Alice's qubits and |ψ⟩₂ and |ψ⟩₃ are Bob's qubits.
    operation CreateEntangledState (qs : Qubit[]) : Unit {
        // Hint: Can you represent this state as a combination of Bell pairs?
        // ...
        fail "Task 2.1 not implemented yet";
    }

Reference Implementation
   // Task 2.1. Entangled state
    operation CreateEntangledState_Reference (qs : Qubit[]) : Unit is Adj {
        // The desired state is equivalent to two Bell pairs split between Alice and Bob.
        for (i in 0..1) {
            H(qs[i]);
            CNOT(qs[i], qs[i + 2]);
        }
    }

The issue: 
If I do 2 H gates and 2 CNOT gates on |0000> I get this as a result: 
//     |ψ⟩ = ((|0⟩₀ ⊗ |0⟩₂ + |1⟩₀ ⊗ |1⟩₂) / sqrt(2)) ⊗ ((|0⟩₁ ⊗ |0⟩₃ + |1⟩₁ ⊗ |1⟩₃) / sqrt(2)),
which should be equivalent to (if I do the math correctly with Dirac notation)
//     |ψ⟩ = ((|+⟩₀ ⊗ |+⟩₂) / sqrt(2)) ⊗ ((|+⟩₁ ⊗ |+⟩₃) / sqrt(2)),


Would you agree?
If not, could you help me understanding the math?

Thank you 
Markus

tcNickolas:
Looks like the two files that are left are the ones that we decided not to change, so they didn't show up as conflicts and are still in the PR. In principle you can fix that by undoing the changes, but closing this PR and opening a new one might end up being faster.

I like your idea of hidden matrix representation - it combines the best of the both approaches :-)

Give me a moment to write out the math for CreateEntangledState in a separate comment...

tcNickolas:
Let's look at just qubits 0 and 2 - they are not entangled with 1 and 3, so we can do the math for them and handle 1 and 3 in exactly the same way.

You're right that after H and CNOT they are in state (|0⟩₀ ⊗ |0⟩₂ + |1⟩₀ ⊗ |1⟩₂) / sqrt(2).

However, the state |+⟩₀ ⊗ |+⟩₂ differs from that state: if you replace each |+⟩ with (|0⟩ + |1⟩) / sqrt(2) and open the brackets, you'll get (|0⟩₀ ⊗ |0⟩₂ + |0⟩₀ ⊗ |1⟩₂ + |1⟩₀ ⊗ |0⟩₂ + |1⟩₀ ⊗ |1⟩₂) / 2, which differs from the state you prepared by the two middle terms |0⟩₀ ⊗ |1⟩₂ + |1⟩₀ ⊗ |0⟩₂ (and the sqrt(2) normalizing coefficient). To get rid of them, you need to add (|0⟩₀ ⊗ |0⟩₂ - |0⟩₀ ⊗ |1⟩₂ - |1⟩₀ ⊗ |0⟩₂ + |1⟩₀ ⊗ |1⟩₂) / 2, which is exactly |-⟩₀ ⊗ |-⟩₂.

Alternatively, you can express |0⟩ = (|+⟩ + |-⟩) / sqrt(2) and |1⟩ = (|+⟩ - |-⟩) / sqrt(2) and open the brackets in (|0⟩₀ ⊗ |0⟩₂ + |1⟩₀ ⊗ |1⟩₂) / sqrt(2) to see that it's exactly the state we need to prep.

tcNickolas:
@mbuchberger1967  I have finished editing Superposition workbook part 2 and merged it to master, so you should be unblocked for adding explanation for Hardy state prep task. The new version of the workbook conflicts with this PR again, so it's definitely better to open a new PR with that change off the latest master than to try and unentangle this one :-)

PR #307, opened 2020-03-19
Mbu/teleportation

mbuchberger1967:
Teleportation/ReferenceImplementation.qs is the relevant change: 
Teleportation Task 4.1. Entangled trio - added comment to solution

PR #308, opened 2020-03-19
GroversAlgorithm - Task 2.2 Explained Conditional Phase flip detailed

mbuchberger1967:
GroversAlgorithm/GroversAlgorithm.ipynb is the relevant change: 
Task 2.2 Explained Conditional Phase flip detailed


tcNickolas:
We agreed to do this change as part of #307, so I'll close this PR. Thank you!

PR #310, opened 2020-03-21
Remove deprecated functions

vivanwin:


RippleCarryAdder/Tasks.qs
```diff
@@ -8,6 +8,7 @@ namespace Quantum.Kata.RippleCarryAdder {
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Convert;
     open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Logical;
```
tcNickolas: This line shouldn't be necessary - looks like the switch only happens in the tests and not in the reference implementation, so the learner won't use the new namespace.

PR #311, opened 2020-03-24
Quantum kata improvements measurement

delbert:
Improving Task 1.5 for demo for Quantum Community


.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
+{
```
tcNickolas: Could you please exclude this file and the tasks.json? Each kata has their own, and it can be confusing to add such files to the root folder, since they are configured to call BasicGates kata specifically.

delbert: ok, i think i got rid of lanch.json and tasks.json from my branch.  i will go through the rest of the Kata and make the other improvements

Measurements/ReferenceImplementation.qs
```diff
@@ -179,10 +185,13 @@ namespace Quantum.Kata.Measurements {
...
 
     operation SuperpositionMeasurement_Reference (qs : Qubit[], bits1 : Bool[][], bits2 : Bool[][]) : Int {
+        Fact(Length(bits1[0]) == Length(bits2[0]), "Second dimension of bit Arrays should have the same length");
+        Fact(Length(bits1[0]) == Length(qs), "Second dimension of bit Arrays should be euqal to number of qubits");
+
         // measure all qubits and check in which array you can find the resulting bit string
         let meas = ResultArrayAsBoolArray (MultiM(qs));
         for (i in 0 .. Length(bits1) - 1) {
-            if (AreBitstringsEqual(bits1[i], meas)) {
+            if (AreBitstringsEqual_Reference(bits1[i], meas)) {
```
tcNickolas, line 194: This can be replaced by a new library function [EqualA](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arrays.equala) (needs an `open` directive) to make the code shorter:

```suggestion
            if (EqualA(EqualB, bits1[i], meas)) {
```

Measurements/ReferenceImplementation.qs
```diff
@@ -371,7 +383,7 @@ namespace Quantum.Kata.Measurements {
 
         // Apply diag(..) (H ⊗ H) diag(..)
         within {
-            ApplyDiag(qs);
+            ApplyDiag_Reference(qs);
```
tcNickolas, line 386: Theoretically this could also be implemented using a library operation, but I'm not sure using it will be more readable (and more instructive!) for the learner. What do you think?

```suggestion
            ApplyDiagonalUnitary([PI(), 0.0, 0.0, 0.0], LittleEndian(qs));
```

Measurements/Tasks.qs
```diff
@@ -124,6 +125,9 @@ namespace Quantum.Kata.Measurements {
     // Example: for bit strings [false, true, false] and [false, false, true]
     //          return 0 corresponds to state |010⟩, and return 1 corresponds to state |001⟩.
     operation TwoBitstringsMeasurement (qs : Qubit[], bits1 : Bool[], bits2 : Bool[]) : Int {
+        Fact(Length(bits1) == Length(bits2), "Bit Arrays should have the same length");
```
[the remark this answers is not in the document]
delbert: makes sense, Mariia.  i removed the FactIO statements from Taks.qs

Measurements/ReferenceImplementation.qs
delbert: hi Mariia, i found Microsoft.Quantum.Arrays >> EqualA, but not EqualB.  if i just use EqualA in my code, i get a compile error, because the function is expecting a type:  ((('T, 'T) -> Bool), 'T[], 'T[]).  so i am not sure how to resolve this.  if there is a definition of EqualB in your code, should i see it in the compare window, here?

Measurements/ReferenceImplementation.qs
delbert: i like this suggestion.  but if i hit "commit suggestion" i am going to leave the ApplyDiag_Reference orphaned in the code, right?  so i should change locally and push changes up, correct?

tcNickolas: Yes, you'll need to do corresponding changes locally, suggestions don't play nicely when the same suggestion applies in multiple places or needs matching changes elsewhere.

Measurements/ReferenceImplementation.qs
tcNickolas: It's [the library EqualB](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.logical.equalb), from M.Q.Logical namespace, that compares booleans.

delbert:
i think i have all of the suggestions incorporated, at this point.

PR #318, opened 2020-04-07
BasicGates katas add clarification about endiannes and qubit order.

bpatra:
After finishing `MultiQubitSystems` and `MultiQubitGates` tutorials and tackling Part II of `Basic Gates` kata, I thought the little-endian notation would be de-facto notation for all katas. For instance this is the notation used in the `DumpMachine` instruction.

I believe that mentioning big-endian is the notation used in this kata could be useful and avoid some confusion from learners.

### DIFF
--- a/BasicGates/BasicGates.ipynb
+++ b/BasicGates/BasicGates.ipynb
@@ -379,6 +379,8 @@
    "source": [
     "## Part II. Multi-Qubit Gates\n",
     "\n",
+    "The Dirac notations in the following use big-endian format (the most significant bit is stored first), see this [tutorial section](../tutorials/MultiQubitSystems/MultiQubitSystems.ipynb#Endianness) for a refresher on endianness.\n",
+    "\n",
     "### Q# materials\n",
     "\n",
     "* Using controlled and adjoint versions of gates is covered in the Q# documentation on [operation types](https://docs.microsoft.com/quantum/language/type-model#operation-and-function-types)."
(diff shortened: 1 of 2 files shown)


bpatra:
You are right, your correction is absolutely what I wanted to express. I'll make the changes.

bpatra:
Actually, I realized that in Basic Gates katas  Test outputs do not print qubit with little-endian notation at al. Therefore, I rewrote without mentioning endianness at all, I hope this will avoid confusing for the newcomers like me.

I reused a similar sentence than the one that is mentioned in [Measurement kata](https://github.com/microsoft/QuantumKatas/blob/master/Measurements/Measurements.ipynb#Task-1.6.-Distinguish-four-basis-states.)

> In this task and the subsequent ones the order of qubit states in task description matches the order of qubits in the array (i.e.,  |10⟩  state corresponds to qs[0] in state  |1⟩  and qs[1] in state  |0⟩ ).

tcNickolas:
If we run task 2.1, we'll get the following output:

image

The task promises a superposition of basis states 00 and 10, and the test prints a state with non-zero amplitudes of 0 and 1, so is seems there is some space for confusion?

(I'm happy to merge your note in its current state, but want to get to the root of the the doubts you had)

bpatra:
Ah ok I got a different output (see code below). I made my fix on an outdated version and then rebased for PR. My bad.

Screenshot 2020-04-13 at 16 45 50

I'll PR again.

bpatra:
@tcNickolas hope that 254c145 is finally what we wanted 😅

PR #319, opened 2020-04-08
Update to Jupyter Core 1.3.

cgranade:
This PR updates the dependency from the Microsoft.Quantum.Katas package to Microsoft.Jupyter.Core to version 1.3, and uses new functionality in that version to provide extended documentation about the `%kata` and `%check_kata` magic commands.

### DIFF
--- a/utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
+++ b/utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Jupyter.Core;
 using Microsoft.Quantum.IQSharp;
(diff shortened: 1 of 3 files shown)


tcNickolas:
Could you please clarify 1) when does this change take effect? (I'm assuming with the next release?) and 2) what does the effect of this change look like? I'm not sure how the documentation for magic commands can be accessed.

cgranade:
> Could you please clarify 1) when does this change take effect? (I'm assuming with the next release?) …

This was the content of https://github.com/microsoft/iqsharp/pull/113 and https://github.com/microsoft/iqsharp/pull/114, and should hopefully be a minor enhancement for the next release of IQ#, yes.

> …and 2) what does the effect of this change look like? I'm not sure how the documentation for magic commands can be accessed.

Since IQ# 0.5, `?` commands can be used to look up documentation on magic commands:

image

This needs some improvement to make use of the logic in https://github.com/microsoft/iqsharp/pull/114, but in the meantime https://github.com/microsoft/iqsharp/pull/113 generates Markdown documents from each magic symbol as a part of builds; these Markdown files could in principle be added to reference documentation.

cgranade:
Thanks for the review, and my apologies for not providing more context from the start!

PR #320, opened 2020-04-11
Fix updateQDKversion.ps1 to account for beta versions

tcNickolas:
In https://github.com/microsoft/QuantumKatas/pull/319 I noticed a leftover `-beta` version in one of our `.csproj` files and realized that our update script would miss it, since the version regex it uses doesn't account for such versions. This change should fix this.

### DIFF
--- a/scripts/updateQDKVersion.ps1
+++ b/scripts/updateQDKVersion.ps1
@@ -23,7 +23,10 @@ param(
 
 $katasRoot = Join-Path $PSScriptRoot "\..\"
 
-$csString = 'PackageReference Include=\"Microsoft\.Quantum\.[a-zA-Z\.]+\" Version=\"(?<oldVersion>[0-9|\.]+)\"'
+# old version regular expression accounts for possible suffixes like "-beta"
+$versionRegex = "(?<oldVersion>[0-9|\.|\-|a-z]+)"
+
+$csString = 'PackageReference Include=\"Microsoft\.Quantum\.[a-zA-Z\.]+\" Version=\"' + $versionRegex + '\"'
 $csFiles = (Get-ChildItem -Path $katasRoot -file -Recurse -Include "*.csproj" | ForEach-Object { Select-String -Path $_ -Pattern "Microsoft.Quantum" } | Select-Object -Unique Path)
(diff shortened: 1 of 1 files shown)


scripts/updateQDKVersion.ps1
```diff
@@ -36,7 +39,7 @@ $csFiles | ForEach-Object {
     } | Set-Content -Encoding UTF8 $_.Path
 }
 
-$ipynbString = '%package Microsoft.Quantum.Katas::(?<oldVersion>[0-9|\.]+)'
+$ipynbString = '%package Microsoft.Quantum.Katas::' + $versionRegex
```
cgranade, line 42: Minor, but you should be able to use string interpolation here to avoid string concatenation.
```suggestion
$ipynbString = "%package Microsoft.Quantum.Katas::$versionRegex"
```

scripts/updateQDKVersion.ps1
```diff
@@ -51,7 +54,8 @@ $ipynbFiles | ForEach-Object {
         } | Set-Content $_.Path
     }
 }
-$dockerString = 'FROM mcr.microsoft.com/quantum/iqsharp-base:(?<oldVersion>[0-9|\.]+)'
+
+$dockerString = 'FROM mcr.microsoft.com/quantum/iqsharp-base:' + $versionRegex
```
cgranade, line 58: Similarly here and at L70.

scripts/updateQDKVersion.ps1
tcNickolas: I tried to use string interpolation with csproj line, and it didn't work, so I decided I got the wrong syntax and dropped the idea...  Do you know why this works in three strings but not the fourth one?

cgranade: I think this is a difference between `'` and `"`:
```PowerShell
PS tmp> $x = "foo"
PS tmp> "${x}bar"
foobar
PS tmp> '${x}bar'
${x}bar
```

tcNickolas: No, with `'` it runs but doesn't give the result I want, but with `"` it threw an exception. Something more interesting I think...

Issue #314, opened 2020-03-30
Alternative solution to Measurements/Hardy State


wsgac:
Hi,

I was fooling around with an approach slightly different from the one in the reference implementation and was able to come up with the following:
```
operation Hardy_State (qs : Qubit[]) : Unit {
    Ry(2.0*ArcCos(Sqrt(10.0/12.0)), qs[0]);
    (ControlledOnInt(1, H))([qs[0]], qs[1]);
    (ControlledOnInt(0, Ry))([qs[0]], (2.0*ArcCos(Sqrt(9.0/10.0)), qs[1]));
}
```
Feel free to include this as an alternative and to comment if this approach has any downsides.

tcNickolas:
*I'm assuming you mean Superposition/Hardy State task, I don't think we have one about this state in Measurements :-)*

Indeed, you can use H gate here instead of Ry with rotation by PI()/4 freely; those two gates differ only by the second column of their matrix representation (H gate has 1 in the first row and -1 in the second row, and Ry(PI()/4) has -1 in the first row and 1 in the second row), but since you only apply them to |0> state, this difference has no effect. Using Ry is more general approach, since it allows to split the |0> state into any superposition of |0> and |1>, as long as they both have real amplitudes.

It's probably not worth including it as a separate solution (since only 1 gate is different), but might be worthwhile to leave a comment in the solution itself, mentioning that you can use H instead of Ry here. Let me know if you'd like to do that!

(Sorry for the delay replying! I wanted to finish https://github.com/microsoft/QuantumKatas/pull/282 first to reduce the number of threads about Superposition solutions)

mbuchberger1967:
Hi,
I agree that in this special case we could use a Controlled H instead of a Controlled Ry, but the maths I just added to this task are guiding the reader to the Controlled Ry gate. 
This is why I changed the reference impl to use a Controlled Ry, and just left a comment that in this special case one could also use a Controlled h. 
I hope you agree.
BR
Markus

tcNickolas:
https://github.com/microsoft/QuantumKatas/commit/8c59741736d8b5a979f16bc88eacae620849635c added a comment to the solution mentioning that Controlled H is an option in this solution, so I think we can close this issue.

Issue #313, opened 2020-03-28
Math expressions would not render correctly when Jupyter notebook used locally in VS Code.


jkovzels:
Most likely, it is a problem of VS Code implementation of Jupyter but math expression with `\\` line breaks will render with **$$** and ruin the at block unless closing `&&` moved to the new line.

here is an example from Complex Arithmetic:
![image](https://user-images.githubusercontent.com/351671/77834903-9f27ef80-7148-11ea-9d8c-87516d676da6.png)

tcNickolas:
I'll admit I've never tried using the Jupyter Notebooks in VS Code, so it will take me a while to reproduce the issue. Thank you for reporting it!

If you have found a simple fix that renders the notebooks better in VS Code (without breaking them in regular Jupyter!), would you be willing to send a pull request with the fix?

jkovzels:
I'll admit that after digging deeper into it I can recommend not to use  Jupyter Notebooks in VS Code :). 
Experience is not good, notebook does not look good, fonts, line length, local links not working, an overall jerkiness. I would recommend people to run the notebooks locally in the browser at this point.

That said, workaround is to regex replace all `([^"\\n])\$\$` with `\1\\n\$\$`: all `$$` with leading character except if character is " or new line, replace with leading char (group `\1`) and `$$`.
But I do not thing that anyone should do that. Fix should come from VS Code Python plugin team.

tcNickolas:
Thank you for the investigation! This is very helpful - now I know not to spend too much time on this :-)

PR #322, opened 2020-04-14
Added maths and solution for Hardy State (task 12)

mbuchberger1967:
Hi @tcNickolas ,
as we agreed I re-did my changes for task 12 "Hardy State", and moved the detailed "matrix and vector calculations" into <details> sections.
But I noticed the following issues:
1. In my Jupyter Notebook (installed with Anaconda) the <details> tag does not work, I cannot expand/collapse the section
2. In the VS Code Notebook editor, in the details sections the interpretation of $...$ to write latex doesnt work, only $$...$$. Do you have a trick to solve this or have another idea?
Please check "Need a hint" sections on step 1, and step 2.a.
In my Notebook editor the details section cannot be collapsed, but the latex works also with $...$

3. In the current reference implementation a "Controlled H" is used for the second rotation. According to your original post and also to my maths I'll end up in using a ControlledOnInt(1) Ry operation.
Due to my understanding this is correct, and cannot be generally replaced by a Controlled H, because they differ in a sign.... in the special case here they bring the same result, so I left it as a comment in the code. 
but in my opinion I would remove this completely, because I would have become confused to be honest.

BR
Markus


tcNickolas:
1. This depends mostly on the browser - I'm using Firefox and things collapse and expand nicely (including in your comment on this pull request :-)), if I recall correctly IE and Edge might have issues recognizing the tag but Chrome processes it fine.

2. I've seen this or similar issue reported in https://github.com/microsoft/QuantumKatas/issues/313 - I think our conclusion there was that VS Code support for Notebooks is not quite polished yet, so we're not investing in that experience, our primary focus is on browser experience.

3. Yes, I used Ry there mostly for uniformity - the learner will have to get comfortable with Ry anyways, at this point they won't be able to get around it with just H gates. I think a comment in the solution will be a reasonable compromise with #314.

I'll look at the change itself shortly (well, before the weekend... :-))

PR #326, opened 2020-04-19
Task 1.7 error in solution description step2

mbuchberger1967:
Step 2 in solution of task 1.7 is wrong, I fied it with this commit.


mbuchberger1967:
hi @tcNickolas 
I fixed the solution description in a simple way. It is now correct in general, and not only related to the sample that was provided.
This was not much effort, and I propose to commit this, to have at least a correct description.

BR
Markus

Issue #243, opened 2019-11-23
(Harmless) Failures when Building the Katas docker image


vxfield:
I'm not sure if this is by-design, but while building the Katas docker image, it produces a couple warnings and several failures in the NbConvertApp markdown conversion.

Although they seem harmless and don't impede the generation of the docker image, it gives the impression that something went wrong in the process.

**Steps to repro:**
1) Run "docker build -t katas ." at the root of the repo, as suggested in the README.

**Warning example:**
```
Step 6/33 : RUN pip install "pytest"
 ---> Running in dc47a6915564
WARNING: The directory '/home/jovyan/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.

```

**Potential solutions to these warnings:**
A) Use the -H flag in sudo.
B) Install pytest in the user's folder without sudo (pip install --user pytest).

**Failure example:**
```
Step 23/33 : RUN ${HOME}/scripts/prebuild-kata.sh SuperdenseCoding
 ---> Running in 6e5cf347f452
Prebuilding: SuperdenseCoding.ipynb in SuperdenseCoding kata...
Microsoft (R) Build Engine version 16.3.0+0f4c62fea for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 57.32 ms for /home/jovyan/SuperdenseCoding/SuperdenseCoding.csproj.
  Restore completed in 442.28 ms for /home/jovyan/SuperdenseCoding/SuperdenseCoding.csproj.
  SuperdenseCoding -> /home/jovyan/SuperdenseCoding/bin/Debug/netcoreapp3.0/SuperdenseCoding.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:18.75
[NbConvertApp] Converting notebook SuperdenseCoding/SuperdenseCoding.ipynb to markdown
[NbConvertApp] Executing notebook with kernel: iqsharp
fail: Microsoft.Quantum.IQSharp.Snippets[0]
      QS6307: Not all code paths return a value.
fail: Microsoft.Quantum.IQSharp.Snippets[0]
      QS6307: Not all code paths return a value.
[NbConvertApp] Writing 6117 bytes to SuperdenseCoding/SuperdenseCoding.md
Removing intermediate container 6e5cf347f452
 ---> 803b8c2641bc
```

[...]

PR #333, opened 2020-04-28
[Draft] New pre-built image for Katas to improve startup

vxfield:
This PR adds a new Dockerfile to create a fully baked Katas image that removes online Nuget sources to improve Package loading times (and therefore the first cell execution of all katas).


build/images/katas/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    ./scripts/prebuild-kata.sh UnitaryPatterns && \
+    ./scripts/prebuild-kata.sh tutorials/ComplexArithmetic ComplexArithmetic.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/ExploringDeutschJozsaAlgorithm DeutschJozsaAlgorithmTutorial.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/ExploringGroversAlgorithm ExploringGroversAlgorithmTutorial.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/LinearAlgebra LinearAlgebra.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/MultiQubitGates MultiQubitGates.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/MultiQubitSystems MultiQubitSystems.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/Qubit Qubit.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/SingleQubitGates SingleQubitGates.ipynb && \
+    rm $HOME/NuGet.config && \
```
cgranade, line 53: I thought that the `NuGet.Config` file for Linux normally had a capital `C` and was typically under `~/.nuget/`?

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    echo "<?xml version=\"1.0\" encoding=\"utf-8\"?>\
```
cgranade, line 54: A comment explaining why this is needed may help.

Dockerfile
```diff
@@ -1,51 +1 @@
...
-# RUN ${HOME}/scripts/prebuild-kata.sh UnitaryPatterns
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/ComplexArithmetic ComplexArithmetic.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/ExploringDeutschJozsaAlgorithm DeutschJozsaAlgorithmTutorial.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/ExploringGroversAlgorithm ExploringGroversAlgorithmTutorial.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/LinearAlgebra LinearAlgebra.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/MultiQubitGates MultiQubitGates.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/MultiQubitSystems MultiQubitSystems.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/Qubit Qubit.ipynb
-RUN ${HOME}/scripts/prebuild-kata.sh tutorials/SingleQubitGates SingleQubitGates.ipynb
+FROM mcr.microsoft.com/quantum/katas:0.11.2004.2715-beta
```
rmshaffer, line 1: Should the change to this Dockerfile happen separately, as a later commit, after this `mcr.microsoft.com/quantum/katas` image has been published to mcr? i.e., if we commit this change now, would it break Binder because the image doesn't exist?

Dockerfile
```diff
@@ -1,51 +1 @@
-# We use the iqsharp-base image, as that includes
```
cgranade, line 1: For use on Binder in practice, we should note updates to the katas image won't be picked up without a corresponding commit in this repo. Adding an action that makes that commit would be a good addition after bringing this PR in.

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+# Install Python dependencies for the Python visualization and tutorial notebooks
+RUN pip install -I \
+        matplotlib \
+        pytest
+
+USER ${USER}
+
+# Pre-exec notebooks to improve first-use start time
+RUN cd $HOME && \
```
rmshaffer, line 25: Consider using `pushd` and `popd` instead of `cd`? Also consider `${HOME}` for consistency with the rest of the file?

build/images/katas/Dockerfile
rmshaffer, line 53: `${HOME}`, for consistency with the rest of the file?

Dockerfile
vxfield: Yes, we will need to do a commit anyway to point to the release version/tag.

build/images/katas/Dockerfile
vxfield: This is a local NuGet.config in the root of the repo that refer to the quantum alpha source.
Since we want to remove all online sources, we need to remove the project's NuGet.config and the user's NuGet.config (that's done below).

build/images/katas/Dockerfile
vxfield: Agree.

Dockerfile
vxfield: Yes, agree. It should be as two separate phases and since I manually pushed it from my local machine, I didn't have a problem.
But it would become an issue when we start automating this...

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,67 @@
+# We use the iqsharp-base image, as that includes
```
tcNickolas, line 1: Could we be any chance move this file to `scripts` folder? As is, this change would create a `build` folder with exactly one script, and it makes sense to have all scripts in the same place.

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+# Make sure the contents of our repo are in ${HOME}
+# Required for mybinder.org
+COPY . ${HOME}
+
+USER root
+RUN chown -R ${USER} ${HOME} && \
+    chmod +x ${HOME}/scripts/*.sh
+
+# Install Python dependencies for the Python visualization and tutorial notebooks
+RUN pip install -I \
```
tcNickolas, line 18: The `-I` switch is new compared to the existing Dockerfile - does it fix the warning reported in https://github.com/microsoft/QuantumKatas/issues/243?

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+        matplotlib \
+        pytest
+
+USER ${USER}
+
+# Pre-exec notebooks to improve first-use start time
+RUN cd ${HOME} && \
+    ./scripts/prebuild-kata.sh BasicGates && \
```
tcNickolas, line 26: Looking at the bigger picture of this PR and #335, what happens (once these two changes are in) when we update one of the existing katas or notebooks or add a new one (such as the QFT kata right now)? Are we planning to build a "released" version of the katas with the release which is going to be the base, and then any changes after that are rebuilt by Binder when a specific kata is executed? Will the new kata have access to Nuget packages downloaded by prebuilding other katas?

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+    ./scripts/prebuild-kata.sh tutorials/LinearAlgebra LinearAlgebra.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/MultiQubitGates MultiQubitGates.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/MultiQubitSystems MultiQubitSystems.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/Qubit Qubit.ipynb && \
+    ./scripts/prebuild-kata.sh tutorials/SingleQubitGates SingleQubitGates.ipynb && \
+# To improve performance when running the %package commands (usually Katas' first cell)
+# we remove all online sources for NuGet such that IQ# Package Loading and NuGet dependency
+# resolution won't attempt to resolve packages dependencies again (as it was already done
+# during the prebuild steps above).
+# The downside is that online packages that were already downloaded to .nuget/packages folder
```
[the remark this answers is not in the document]
vxfield: Yes. I'm updating the comment to "The downside is that only online packages that were already downloaded"

build/images/katas/Dockerfile
```diff
@@ -0,0 +1,67 @@
+# We use the iqsharp-base image, as that includes
+# the .NET Core SDK, IQ#, and Jupyter Notebook already
+# installed for us.
+FROM mcr.microsoft.com/quantum/iqsharp-base:0.11.2004.2715-beta
```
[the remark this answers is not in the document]
vxfield: Yes. I'll update the ps1 script to keep this Dockerfile updated as well.

build/images/katas/Dockerfile
vxfield: Since this is technically not a script, I though it would be better to have under a generic build folder, under images.
I think we should then move the scripts folder under the build folder (as it is with the other repos), but I didn't have time to do it yet (as I would need to update the QDK build)

build/images/katas/Dockerfile
vxfield: No, this is to overwrite previous installed packages.
-I, --ignore-installed
Ignore the installed packages, overwriting them. This can break your system if the existing package is of a different version or was installed with a different package manager!

#243 is still happening.

build/images/katas/Dockerfile
vxfield: Great question that requires a longer answer.

**About the image**

With a Katas image, it will only contain what was available when that image was built, so we would have a monthly Katas release aligned with QDK.
Whatever changes happened between when the image was built and the next release won't be available in MyBinder, as it would normally reuse the pre-built image the the "official" release.

Some options to consider:
1) We could also have another branch where MyBinder will pick the latest changes in master (even before the monthly QDK release), but that would require some automation to keep that branch in sync with master.
2) Alternatively, we could have #335 in a special branch that is used by MyBinder to fetch the pre-built image instead of it being the default mode. I think this is what @cgranade did for the samples.

Now I think that option 2 is likely the best one as it keeps the current Katas behavior but offers an option for an optimized MyBinder experience.

**About the Nuget packages**

About the Nuget packages, the pre-built image will pre-download all packages referenced by all Katas at the moment of the build, and then it disable online packages (so if users try to load packages that were not originally referenced by any Katas, that package won't be loaded). 

Alternatively, we could allow the pre-built image to fetching online packages, and only remove that option for the optimized MyBinder experience with #335. 

**Conclusion**

With that, I think it's best to leave this original Dockerfile in the root as is (so people can still attempt to run it on MyBinder) and offer an optimized Dockerfile in a separate KatasImage branch that will point to the pre-built image based on the original root Dockerfile.

vxfield:
After @tcNickolas comments, I think it's best to abandon this PR and only have the #335 PR in a special MyBinder branch as @cgranade did for Samples.

tcNickolas:
Are there any fragments of this change that can be applied to the master Dockerfile, to improve the experience people have running from it? I agree that option 2 you described is better (thank you for writing it up in such level of detail!), but it will involve updating all our links to Binder to use that special branch. I can imagine some missed or non-updatable links (such as in tweets), so it would be nice to have a reasonable experience when running Binder from master. It seems that disabling the online NuGet sources after prebuilding them will still yield the speedup on `%package` commands?

vxfield:
Yes, we can disable the online Nuget sources on the current Dockerfile.
Can you check/approve PR #336? After that I'll do another small PR to combine all the RUN commands into one (slightly reduce image size by creating just one layer instead of a dozen) and disable the Nuget online sources.

Regarding updating links, we should stick to using aka.ms for that so we can update it any time. Do we have one for Katas already?

vxfield:
Created new PR #337 that transfer optimizations from this PR without doing a breaking change in MyBinder behavior that this PR would do

PR #338, opened 2020-05-01
Dockerfile that uses pre-build Katas docker image

vxfield:
This PR replaces, in the ⭐binder branch, the original Katas Dockerfile (that starts from iqsharp-base and build all Katas) with a simple Dockerfile that just points to the pre-built Katas docker image.

It can be tested on my fork via
https://mybinder.org/v2/gh/vxfield/QuantumKatas/⭐binder?urlpath=%2Fnotebooks%2Findex.ipynb

The final official URL for MyBinder would be: 
https://mybinder.org/v2/gh/microsoft/QuantumKatas/⭐binder?urlpath=%2Fnotebooks%2Findex.ipynb

We should create/update all of our aka.ms Katas URLs to use this special branch.

**To-do:**
1) Update the Readme section for local Docker execution, pointing users to use the pre-built image instead of building locally.
2) Document the manual steps (for now) that will build and publish the Katas pre-built image to MCR. I'll create a work-item to update the QDK build process to do it automatically.


rmshaffer:
Do you mean to target the `⭐binder` branch with this PR, rather than `master`?

cgranade:
> 2. Document the manual steps (for now) that will build and publish the Katas pre-built image to MCR. I'll create a work-item to update the QDK build process to do it automatically.

Would it work to follow the same process as on the samples repo?

vxfield:
@ all: My bad, I wanted to target the same ⭐binder branch but missed it, and did the PR more as a place for review and discussion (and because I was working on my fork).

@cgranade: yes, it would work the same as it works for Samples.

@tcNickolas: I agree with need to automate this, but it would be a once-in-a-month 5-min-step as part of the QDK release (and done by the person working on the QDK release), so I don't see why we should block this perf improvement on that. Anyway, I think I can probably automated it before the next release.

tcNickolas:
Sorry, you lost me again... From your previous comments my understanding was that if we switch Binder to a pre-built image, we need to update that image every time the Katas are updated, otherwise it will keep using the released version. I don't want the Binder users to use the "released" version of the katas (which is why we go into the whole story of using a separate branch?), so it would not be once a month, it would be once per each Katas master update.

vxfield:
I synced with Mariia offline and (finally!) understood that the requirement is to have the Katas releases decoupled with the monthly QDK releases (we want Katas to be released more regularly).

So we would need an action that every time master changes, we build a new image and update the binder branch.
I think this is what @cgranade started with the Samples: https://github.com/Microsoft/Quantum/actions/runs/93296882/workflow

But anyway, we agreed to postpone this effort as there is little value to end-users (this will only help improve the initial MyBinder downtime ~30 min per release after master has changed).

cgranade:
> But anyway, we agreed to postpone this effort as there is little value to end-users (this will only help improve the initial MyBinder downtime ~30 min per release after master has changed).

For clarity, it can be more than 30 minutes per katas release, as mybinder.org doesn't share caches amongst the different Kubernetes clusters that it load balances onto, so it's about 30 minutes per cluster per release. If a user gets unlucky and is the first to be load balanced onto a cluster, that can cause a problem. I don't know how often mybinder.org caches are cleared either, such that it could be more frequent if caches are cleared between kata releases.

PR #347, opened 2020-05-10
Fix column width for minus i

abienvenu:
In the table telling the composition of standard ket states (|+>, |->, |i>, |-i>) in zero/one basis, the "|-i>" formula is wrapped on two lines, hurting readability (tested with both Firefox and Chromium with Ubuntu 20.04).

In this pull request, I suggest to widen the table columns a little bit, so that each formula will fit on a single line.


vivanwin:
Hi,
Looks good, but maybe we could get rid of the fixed width all together with default markdown tables like this: 

```Markdown
|   |   |
|:---:|:---:|
|$|+\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle + |1\rangle\big)$|$|-\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle - |1\rangle\big)$|
|$|i\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle + i|1\rangle\big)$|$|-i\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle - i|1\rangle\big)$|
```

Not sure if this would also work on your setup.

tcNickolas:
@vivanwin I don't really like how default Markdown tables look when they don't have a header, and in this case in particular I like to have some extra space between formulas in the same row to keep them visually separate:

image

In some cases they look nicer, but in this case I think we'll be better using the HTML formatting.

PR #348, opened 2020-05-11
Switch to latest images for build agents

anjbur:
Switching QDK pipeline to use latest images for build agents. There are corresponding PRs in other repos as well.


tcNickolas:
Do you know why the build takes so long to finish this time? It looks like it's on its third attempt, which is not something I've seen happen previously - I used to have to kick it again. And the last attempt spun off three separate versions of Windows build - is it because it was Windows build that failed on previous attempts?

anjbur:
It seems to be re-starting failed builds and also starting a new one, then canceling all but the new build. Mac also failed the first time, and it has a "canceled" build showing as well. I'm looking into why it's failing, it succeeded when I was testing.

tcNickolas:
Looks like the failure was not caused by the attempt to double-install IQ#: that warning went away but the failure persists. The error looks to be

    At D:\a\1\s\scripts\validate-notebooks.ps1:57 char:9
    +         jupyter nbconvert $CheckNotebook --execute  --ExecutePreproce ...
    +         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        + CategoryInfo          : NotSpecified: ([NbConvertApp] ...k.ipynb to html:String) [], RemoteException
        + FullyQualifiedErrorId : NativeCommandError

anjbur:
Thanks, yeah I've been looking into it. Since the test fails when a notebook would be executed with the IQ# kernel and the IQ# installation is still throwing an error that wasn't there previously, it seems that the actual problem might be with the installation. I'm still looking for a fix.

anjbur:
There shouldn't be any behavioral changes. I got rid of the `$errors` variable, but it wasn't being used for anything (removed it because it was somehow interfering with output being printed if it had been redirected from stderr to stdout). Setting `$ErrorActionPreference = 'Continue'` in parts of the script means that the script won't throw on non-terminating errors there, but since we're checking the return code that shouldn't change whether the script fails or succeeds. Generally, the changes force Powershell to print messages sent to stderr (which is the current behavior on master) rather than throwing on them.

tcNickolas:
Thank you for clarification!

PR #352, opened 2020-05-12
Reorder add extend superposition notebook

vivanwin:


Superposition/Tasks.qs
```diff
@@ -50,21 +50,8 @@ namespace Quantum.Kata.Superposition {
...
-    operation UnequalSuperposition (q : Qubit, alpha : Double) : Unit {
-        // Hint: Experiment with rotation gates from the Microsoft.Quantum.Intrinsic namespace.
-        // Note that all rotation operators rotate the state by _half_ of its angle argument.
-
-        // ...
-    }
-
-
-    // Task 4. Superposition of all basis vectors on two qubits
+
```
tcNickolas, line 53: GitHub doesn't allow me to leave the comment where it belongs: let's define part I above the 1st task to match part II later in the kata, as part of splitting the tasks in two groups. (Similar format to Measurements kata as well)

PR #354, opened 2020-05-19
Set IQ# HostingEnvironment to "build-agent"

vxfield:
This PR sets IQ# HostingEnvironment to "build-agent" in order to identify telemetry coming from CI processes.


vxfield:
I’ll need to update the Telemetry ingestion pipeline, but yes, it will work.

PR #369, opened 2020-06-09
Switch simons algorithm kata to use common counter simulator

joesho112358:


utilities/Common/CounterSimulator.cs
```diff
@@ -34,6 +34,14 @@ public class CounterSimulator : QuantumSimulator
         {
             this.OnOperationStart += CountOperationCalls;
         }
+        
+        /// <summary>
+        /// Getter method for _operationsCount.
```
tcNickolas, line 39: I would clarify this comment to emphasize that this is not just a getter method, it's a method that is designed to be accessed from C# code, as opposed to `GetOracleCallsCount` which is supposed to be called from Q# (this gave me a pause, since I was quite certain that `OracleCounter` already has a method that accesses the number of calls... :-))

joesho112358:
I will try to get to those issues tonight. I guess I should also remove the unused `OracleCounterSimulator.cs` too!

tcNickolas:
Right, I knew there was one more thing I had in mind, removing unused file was exactly it!

utilities/Common/CounterSimulator.cs
joesho112358: Should be updated with the latest

tcNickolas:
I have this review high on my TODO list, just haven't had 30 minutes to sit down and look at it in the past couple of days... I'll try to look tonight or tomorrow. Thank you for your patience!

joesho112358:
No worries- I was as much exploring some of the new actions on Github as I was curious!

joesho112358:
Ah, I didn't know about that. Probably should have thought about it after seeing the CI builds. Thanks though!

tcNickolas:
The build passes even without this change - it's just that the next person who builds the kata will introduce this change and be surprised by it :-)

Issue #196, opened 2019-10-09
Add addition/subtraction modulo 2ᴺ tasks to RippleCarryAdder kata


tcNickolas:
RippleCarryAdder offers great coverage of traditional addition/subtraction operations; operations modulo 2ᴺ are rather similar, reusing a lot of the same primitives, so it makes sense to add a section covering modular operations. Addition modulo some other number would probably be out of scope for this change - if we want to develop that direction, we should branch it off to a separate kata.

jainvasu631:
I am slightly confused by this issue. The RippleCarryAdder kata has N-bit Adder (Task 1.7 and Task 2.5) and N-bit subtractor (Task 4.1). These are in effect implementations of Addition/Subtraction modulo 2^n. 
@tcNickolas can you clarify what tasks you wish to be added when say operations modulo 2^n

tcNickolas:
The existing tasks cover the normal addition, which ends up having a carry qubit to store the (N+1)-th bit of the sum (most significant). In classical computing, you would've been able to just discard this carry qubit to get the result modulo 2ᴺ, but in quantum computing you can't erase information that easily, so you need to modify the computation itself so that the last carry qubit is not computed.

jainvasu631:
Ah I see, what you mean. The task should entail either not using the carry qubit or computing it and then undoing the computation to restore it to its previous value. 

I would like to take this up.

tcNickolas:
Sounds good! (Meanwhile I'll try to catch up on the rest of your PRs :-))

PR #395, opened 2020-07-04
[RippleCarryAdder] Modulo 2ᴺ Tasks Added

jainvasu631:
I added 5 Tasks to RippleCarryAdder in a new section **Part V: Operations Modulo 2ᴺ**. 
These tasks are Addition and Subtraction Modulo N both normal and inplace variants and also a Two's Complement Task.

This Pull Request is with reference to Issue [#196 - Add addition/subtraction modulo 2ᴺ tasks to RippleCarryAdder kata](https://github.com/microsoft/QuantumKatas/issues/196).


jainvasu631:
@tcNickolas Could you take look at this

jainvasu631:
@tcNickolas 
Could you take a look at this PR.
The Issue #196 which this PR addresses is also almost a year old as it was opened in the last Hacktoberfest.

tcNickolas:
I promise you I have not forgotten about any one of the pending PRs... I am the primary maintainer of this project, with some help from Vincent and Mathias, and I'm doing all I can to try and stay on top of the contributions, but unfortunately this is not the only work I do, and I don't have a lot of time I can spend on the Katas :-( I appreciate your patience!

jainvasu631:
I apologize for my pestering and impatience :(

tcNickolas:
It is indeed taking me too long to get to this PR - and it's you contributing to our project, not the other way around!

I asked @msoeken to take a first look - he should be able to get to it sooner than I will.

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+        ApplyToEachA(X, a);
+        
+        // Do (a+1) mod 2ᴺ
+     
+        X(a[0]); // Increment the least significant bit
+     
+        // Since we are incrementing by One, we flip the next bit only if all previous bits are 0.
+        for(i in 1 .. Length(a)-1){
+            (ControlledOnInt(0,X))(a[0..i-1], a[i]); 
+        }
```
msoeken, line 331:
```suggestion
        // Since we are incrementing by One, we flip the next bit only if all previous bits are 0.
        for (prefix in Prefixes(a)) {
            (ControlledOnInt(0, X))(Most(prefix), Tail(prefix));
        }
```

You could use the `M.Q.Arrays.Prefixes` function to make the pattern more explicit. In either case, the `X(a[0])` can be moved into the `for`-loop if you range over all indexes. A controlled operation on an empty array of controls will always be executed.

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+    }
+    
+    // Task 5.3. Subtractor Modulo 2ᴺ
+    operation SubtractorModuloNbits_Reference (a : Qubit[], b : Qubit[], diff : Qubit[]) : Unit is Adj {
+        // Transform a into its Two's Complement 2ᴺ - a 
+        TwosComplement_Reference(a);
+        
+        // Add 2ᴺ - a and b to get (2ᴺ + b - a) mod 2ᴺ = (b-a) mod 2ᴺ 
+        AdderModuloNbits_Reference(a,b,diff);
```
msoeken, line 340:
```suggestion
        AdderModuloNbits_Reference(a, b, diff);
```

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+        
+        let N = Length(a);
+        
+        using (tempCarry = Qubit()) {
+            let carries = [tempCarry] + a;
+
+            // Compute carry bits
+            for (i in 0 .. N-1) {
+                Majority_Reference(a[i], b[i], carries[i]);
+            }
```
msoeken, line 358:
```suggestion
            ApplyToEachA(Majority_Reference, Zipped3(a, b, carries));
```

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+
+            // No need to save last (N+1)th carry bit as Addition is Modulo 2ᴺ
+            
+            // Restore inputs and ancilla, compute sum
+            for (i in N-1 .. -1 .. 0) {
+                UnMajorityAdd_Reference(a[i], b[i], carries[i]);
+            }
```
msoeken, line 365:
```suggestion
            ApplyToEachA(UnMajorityAdd_Reference, Reversed(Zipped(a, b, carries)));
```

If you apply these changes the declaration of `N` can be removed.

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+        }
+    }
+    
+    // Task 5.5. In Place Subtractor Modulo 2ᴺ
+    operation InPlaceSubtractorModuloNbits_Reference (a : Qubit[], b : Qubit[]) : Unit is Adj {
+        // Notice that Task 5.4 is actually the Adjoint of Task 5.3. 
+        // Task 5.3 maps (a,b) -> (a,(a+b)mod2ᴺ). Let c = (a+b)mod2ᴺ
+        // Task 5.4 maps (a,b) -> (a,(b-a)mod2ᴺ). So Task 5.4 will maps (a,c) -> (a,(c-a)mod2ᴺ) = (a,((a+b)mod2ᴺ-a)mod2ᴺ) = (a,b).
+        Adjoint InPlaceAdderModuloNbits_Reference (a,b);
```
msoeken, line 374:
```suggestion
        Adjoint InPlaceAdderModuloNbits_Reference(a, b);
```

RippleCarryAdder/Tasks.qs
```diff
@@ -342,4 +342,77 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+    // Part V. Operations Modulo 2ᴺ
+    //////////////////////////////////////////////////////////////////
+
+    // Through Parts I-IV we have used an extra qubit to act as a carry or borrow qubit. 
+    // These tasks cover the normal addition, which ends up having a carry qubit to store the (N+1)-th bit of the sum (most significant). 
+    // In classical computing, we can easily discard this carry qubit to get the result modulo 2ᴺ. 
+    // However in quantum computing information cannot be erased that easily. 
+    // This extra qubit is left changed after the operation. 
+    // We need to modify the computation itself so that the last carry qubit is not computed or not used at all.
+    // In this part we will now try to implement Operations which either do not use this extra qubit as input.
```
msoeken, line 357:
```suggestion
    // In this part we will now try to implement operations which either do not use this extra qubit as input.
```

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -288,4 +288,89 @@ namespace Quantum.Kata.RippleCarryAdder {
         // transform 2ᴺ - 1 - (b - a) into b - a
         ApplyToEachA(X, b);
     }
+    
+    //////////////////////////////////////////////////////////////////
+    // Part V. Operations Modulo 2ᴺ
+    //////////////////////////////////////////////////////////////////
+
+    // Task 5.1. Adder Modulo 2ᴺ
+    operation AdderModuloNbits_Reference (a : Qubit[], b : Qubit[], sum : Qubit[]) : Unit is Adj {
```
msoeken, line 297: You could add an assertion to check that `sum` is in the all-zero state (see `AssertAllZero`)

RippleCarryAdder/Tasks.qs
```diff
@@ -342,4 +342,77 @@ namespace Quantum.Kata.RippleCarryAdder {
...
+    //      1) N qubit register "a" in an arbitrary state |φ⟩,
+    //      2) N qubit register "b" in an arbitrary state |ψ⟩,
+    // Goals:
+    //      1) transform the "b" register into the binary difference of φ from ψ,
+    //      2) Leave register a unchanged.
+    //      3) Any extra qubits used must be reset to the same state at the end.    
+    operation InPlaceSubtractorModuloNbits (a : Qubit[], b : Qubit[]) : Unit is Adj {
+        // ...
+    }
 }
```
msoeken, line 418:
```suggestion
}

```

msoeken (dismissed):
Thanks @jainvasu631, looks good to me. I left some comments.

RippleCarryAdder/ReferenceImplementation.qs
jainvasu631: Good Idea

RippleCarryAdder/ReferenceImplementation.qs
jainvasu631: I think here you meant Zipped3 not Zipped

jainvasu631:
@msoeken Thanks for the review, I have done all the suggested changes.

RippleCarryAdder/ReferenceImplementation.qs
jainvasu631: @msoeken This Assert actually may not be such a good idea. I added it initially as per your suggestions. However this caused the CI Build to Keep Failiing for various reasons. 

The root cause of this failure was that the test calls the operation AssertOperationEqualReferenced. This basically tests if the reference version of the operation and the user implemented version is identical. When running %check_kata it should ideally be comparing the Reference version with itself and thus be identical by definition and not fail. However presence of the AssertAllZero causes AssertOperationEqualReferenced to fail resulting in the test failing. 

Since AssertOperationEqualReferenced tests identical behaviour on all possible inputs , my guess is that when it is testing on some input where sum is not zero the assert fails and hence the test fails. This is also backed up by the fact that the AssertAllZero condition is not used by any of the previous tasks involving Addition/Subtraction.

jainvasu631: @msoeken After undoing this change the build again seems to be working.

tcNickolas (approved):
Apologies for the delay reviewing this - it took until Christmas holidays to find uninterrupted time for the review :-(

I did a round of cleanup, most prominently the tests for tasks 5.1 and 5.3 - they can't rely on `AssertOperationsEqualReferenced`, since they promise that the sum (and the difference) register will be in 0 state, and that assertion considers inputs in arbitrary states. I updated the code to use a modification of Choi–Jamiołkowski isomorphism, same as task 1.7 does.

Looks great, thank you! (And thank you for your patience with the delays!)

Issue #349, opened 2020-05-11
Problems in understanding the solution for Measurements - Task 2.3**. Peres/Wooters game


mbuchberger1967:
Hi @tcNickolas  and all others,
I have big problems in understanding the solution for 2.3 Peres/Wooters Game:

1. M' is not unitary (cols dont have norm = 1)
2. why do you choose 
![grafik](https://user-images.githubusercontent.com/25125667/81611570-90855700-93db-11ea-84e8-fe33e8dc2342.png)
and not orthogonal vectors to A, B, C?
Some how you do this then in the embedding step, I guess??
3. How exactly do you construct  M'? I cannot see how you use M to embed it into M' exactly... It is somehow contained in M', but the columns are shuffeled? Can you explain your approach?
3b. While discussing with someone on stack-exchange he gave me this proposal which seems to be more straight forward to me...

![grafik](https://user-images.githubusercontent.com/25125667/81611644-b14dac80-93db-11ea-963a-d1520c8f9411.png)

![grafik](https://user-images.githubusercontent.com/25125667/81611675-bd396e80-93db-11ea-81a3-3bdc7e32c9f9.png)

see [stackexchange](https://quantumcomputing.stackexchange.com/questions/11810/ms-quantum-katas-measurements-task-2-3-problems-in-understanding-the-solutio)

4. From M' on the math is clear, but not the q# coding. Where do the X(q) and CNOT(a, q) come from? I cannot discover this in your mathematical explanation?

Thank you very much for enlighten me :)

BR
Markus

mbuchberger1967:
I thought I could use Markdown to describe the matrices, I try it an other way....

tcNickolas:
No, GitHub doesn't support LaTeX, you can do the formulas only using Unicode or screenshots.

This problem (and the explanation) was actually written by @martinquantum, so I'm afraid I can't really enlighten you on it. @martinquantum, could you address this question?

mbuchberger1967:
Hi Mariia, thx for the quick response. I thought you would know it as I found this link to a paper where you and martin are named as the authors 
[](https://assets.codeforces.com/rounds/1116/contest-editorial.pdf)
@martinquantum : maybe you can help me with this?

mbuchberger1967:
Hi @tcNickolas , @martinquantum 
I have an additional question. Meanwhile I can understand how you constructed M'. And if we use 1/sqrt(3) instead 1/sqrt(2) the matrix is unitary.
At the end you state:
[...]

PR #396, opened 2020-07-05
[Measurements] Math Error Fixed: Unitary M'

jainvasu631:
M' is a unitary and hence the norm of its columns should be 1. It had been divided by sqrt(2) but actually should have been divided by sqrt(3). This was also mention in Issue #349.  Moreover the 4th column should be -iqsrt(3) instead of -i. I have fixed these error.


tcNickolas:
@martinquantum Could you please take a look at this PR and the linked issue https://github.com/microsoft/QuantumKatas/issues/349? Thank you!

tcNickolas (approved):
Looks good, thank you!

Apologies for the delay reviewing this - I'm not closely familiar with this task so it took me a while... I ended up getting some help from @msoeken and using the [new DumpOperation functionality](https://github.com/microsoft/Quantum/blob/main/samples/diagnostics/dumping/Dumping%20States%20and%20Operations.ipynb) to verify that the sign of the last column in the implementation matches the sign in the operation implemented by the code (the bottom right element can be chosen up to a phase, and I wanted to double-check that it's a match).

Thank you for your patience with us!

PR #398, opened 2020-07-06
QRNG tutorial

anjbur:
Random number generation tutorial.


tutorials/RandomNumberGeneration/README.md
```diff
@@ -0,0 +1,2 @@
+# Welcome!
+
```
tcNickolas, line 2: Here I would add a sentence or two that would repeat the first paragraph of the tutorial (you'll quantum random number generation, which is very simple but potentially useful application, etc.), and brief instructions on running the tutorial (similar to https://github.com/microsoft/QuantumKatas/blob/master/tutorials/ExploringDeutschJozsaAlgorithm/README.md - it's not designed to be used with C# project)

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+    open Microsoft.Quantum.Convert;
+    open Microsoft.Quantum.Math;
+    open Quantum.Kata.Utils;
+    open Microsoft.Quantum.Arrays;
+    
+
+    // Exercise 1.
+    operation T1_RandomBit_Test () : Unit {
+        Message("Testing...");
+        mutable count = 0;
```
tcNickolas, line 22: Running the exercise as it is given in the code stub (returning -1) gives a somewhat unexpected error "Unexpectedly low number of 1's generated. Generated -1000 out of 1000". I would add an extra check to verify that the returned numbers are always in the valid range.

Is there a way to reuse `CheckFlatDistribution` here as well?

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+        return RandomTwoBits();
+    }
+    
+    // Exercise 3.
+    operation T3_RandomNBits_Test () : Unit {
+        Message("Testing N = 1...");
+        CheckFlatDistribution(RandomNBits, 1, 0, 2, 465);
+        Message("Testing N = 2...");
+        CheckFlatDistribution(RandomNBits, 2, 1, 2, 200);
+        Message("Testing N = 10...");
```
tcNickolas, line 52: I would add test for N = 3 (should be rather fast still)

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+        let max = PowI(2, numBits);
+        mutable counts = ConstantArray(max, 0);
+        mutable average = 0;
+        let total = 1000;
+
+        ResetOracleCallsCount();
+        for (i in 1..total) {
+            let val = f(numBits);
+            set average = average + val;
+            set counts w/= val <- counts[val] + 1;
```
tcNickolas, line 71: Here returning -1 gives "Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')" - this error is a bit too implementation-specific, it doesn't directly tell the user that they returned a value out of range. I'd add a separate check for that here as well to give a more direct error message.

tutorials/RandomNumberGeneration/ReferenceImplementation.qs
```diff
@@ -0,0 +1,57 @@
...
+    operation RandomBit_Reference () : Int {
+        using (q = Qubit()) {
+            H(q);
+            return MResetZ(q) == Zero ? 0 | 1;
+        }
+    }
+
+    // Exercise 2. 
+    operation RandomTwoBits_Reference () : Int {
+        let firstBit = RandomBit_Reference();
```
tcNickolas, line 31: I would simplify it to `return RandomBit_Reference() * 2 + RandomBit_Reference();`

tutorials/RandomNumberGeneration/ReferenceImplementation.qs
```diff
@@ -0,0 +1,57 @@
...
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Measurement;
+    open Microsoft.Quantum.Math;
+    
+
+    // Exercise 1.
+    operation RandomBit_Reference () : Int {
+        using (q = Qubit()) {
+            H(q);
+            return MResetZ(q) == Zero ? 0 | 1;
```
tcNickolas, line 25: Release 0.12 allows us to return qubits in |1⟩ state! So we can simplify a bit

```suggestion
            return M(q) == Zero ? 0 | 1;
```

tutorials/RandomNumberGeneration/ReferenceImplementation.qs
```diff
@@ -0,0 +1,57 @@
...
+        } else {
+            return secondBit == 0 ? 2 | 3;
+        }
+    }
+
+    // Exercise 3.
+    operation RandomNBits_Reference (N: Int) : Int {
+        mutable result = 0;
+        for (i in 0..(N - 1)) {
+            set result = result + RandomBit_Reference() * PowI(2, i);
```
tcNickolas, line 44: In my solution I used `set result = result * 2 + RandomBit_Reference();`, which is a bit simpler - up to you if you want to use that version or keep as is, both are fine.

tutorials/RandomNumberGeneration/ReferenceImplementation.qs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        return result;
+    }
+
+    // Exercise 4. 
+    operation WeightedRandomBit_Reference (x : Double) : Int {
+        let theta = ArcCos(Sqrt(x));
+        using (q = Qubit()) {
+            Ry(2.0 * theta, q);
+            return MResetZ(q) == Zero ? 0 | 1;
```
tcNickolas, line 54:
```suggestion
            return M(q) == Zero ? 0 | 1;
```

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+        }
+        CheckRandomCalls();
+
+        set average = average / total;
```
tcNickolas, line 75: I'm not sure whether you mean integer division or floating-point division here? Integer division will round down here, it might be exactly what you intend but wanted to double-check.

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+
+        for (i in 0..max - 1) {
+            if (counts[i] < minimum) {
+                fail $"Unexpectedly low number of {i}'s generated. Only {counts[i]} out of {total} were {i}";
+            }
+        }
+    }
+
+    operation FindMedian (counts : Int [], arrSize : Int, sampleSize : Int) : Int {
+        mutable x = 0;
```
tcNickolas, line 94: I would call this variable something more descriptive, maybe `totalCount`?

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+        CheckXPercentZero(0.5);
+        CheckXPercentZero(0.75);
+        CheckRandomCalls();
+    }
+
+    operation CheckXPercentZero (x : Double) : Unit {
+        Message($"Testing x = {x}...");
+        mutable oneCount = 0;
+        ResetOracleCallsCount();
+        for (N in 1..1000) {
```
tcNickolas, line 119: We'll need a check for return out of range here as well

tutorials/RandomNumberGeneration/Tests.qs
```diff
@@ -0,0 +1,152 @@
...
+            set oneCount += WeightedRandomBit(x);
+        }
+        CheckRandomCalls();
+
+        let zeroCount = 1000 - oneCount;
```
tcNickolas, line 124: `1000` is hardcoded quite a lot in this test - it will be more convenient to define a variable `nRuns = 1000` in the beginning and refer to it throughout the code. Then if you want to change the number of runs, you change it in one place, with less risk of introducing discrepancies.

tutorials/RandomNumberGeneration/Tests.qs
anjbur: It was intended to be integer division, because that's all that was needed with larger N values. I refactored it to be floating-point though, which is better especially for the N=1 tests.

PR #400, opened 2020-07-11
[TruthTables] MinorFix Q# conditional expressions has else as vertical bar '|' not colon ':'.

jainvasu631:
I spotted this in TruthTables Kata. I think C style syntax was used here as most people are familiar with ```if ? then : else``` while Q# has ```if ? then | else```. 
Since Katas are about learning Q# , I believe for consistency we should use ```|``` not ```:```.


jainvasu631:
Sure

PR #407, opened 2020-07-15
Replacing deprecated IconUrl with Icon and updating the assets for NuGet packages.

ricardo-espinoza:


ricardo-espinoza:
Hi @tcNickolas , the reason for the location of that file is that it will be shared across repositories once it's migrated into Solid. Since the /build directory (or /scripts in the case of QuantumKatas) is the target location of the files coming from Solid, the idea is to place them there in each repository, but once they are available as a step in the Solid build, they will all be placed at compile time from a single main copy.

PR #408, opened 2020-07-15
[SolveSATWithGrover] Minor Error Fixes

jainvasu631:
There is no Task 1.7. It should be 3-SAT not 2-SAT.
`Tasks.qs` seems to have these changes already done.


jainvasu631:
> Nice catch! Interesting that Tasks.qs has it right - somehow I introduced those changes when moving the content over...
> 
> Looking at the two versions of those tasks, it looks like Tasks.qs has a [hint](https://github.com/microsoft/QuantumKatas/blob/master/SolveSATWithGrover/Tasks.qs#L202) which was also lost in transition to notebook - would you be willing to add it to the notebook as part of this change? (We'll need to update your PR to latest master anyways and wait for CI to finish...)

I have made the corresponding changes and also updated the branch.

PR #410, opened 2020-07-15
Validate script

anpaz-msft:
Creating a validate.ps1 that validates Katas and it's seperate from the build/test/pack that scripts, so the later can deal exclusively with the projects that deal with the creation of binaries, and not validating the content of katas.


scripts/test.ps1
```diff
@@ -1,15 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-$ErrorActionPreference = 'Stop'
+Write-Host "No unittests defined for this repo."
```
tcNickolas, line 4: This reminds me: we do have unittests in https://github.com/microsoft/QuantumKatas/tree/master/utilities/DumpUnitaryTest. Are we skipping them today? It would be nice to add that validation

scripts/validate.ps1
```diff
@@ -0,0 +1,35 @@
...
+      /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
+      /property:Version=$Env:ASSEMBLY_VERSION `
+      /property:QsharpDocsOutDir=$Env:DOCS_OUTDIR
+
+  $script:all_ok = ($LastExitCode -eq 0) -and $script:all_ok
+}
+
+& "$PSScriptRoot/set-env.ps1"
+
+# Validating all katas projects can be disables with the ENABLE_KATAS flag:
```
tcNickolas, line 25:
```suggestion
# Validating all katas projects can be disabled with the ENABLE_KATAS flag:
```

scripts/validate-projects.ps1
```diff
@@ -0,0 +1,27 @@
...
+      -c $Env:BUILD_CONFIGURATION `
+      -v $Env:BUILD_VERBOSITY `
+      /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
+      /property:Version=$Env:ASSEMBLY_VERSION `
+      /property:QsharpDocsOutDir=$Env:DOCS_OUTDIR
+
+  $script:all_ok = ($LastExitCode -eq 0) -and $script:all_ok
+}
+
+# Validating all katas projects can be disables with the ENABLE_KATAS flag:
```
tcNickolas, line 25: This comment does not seem true any longer? It seems that we dropped `ENABLE_KATAS` flag altogether in favor of the new split between different build scripts

scripts/test.ps1
tcNickolas: We are indeed skipping those tests, and I think it would be nice to add them. `dotnet test utilities\DumpUnitaryTest` runs the test.

scripts/build.ps1
```diff
@@ -25,14 +25,6 @@ function Build-One {
...
-# Building all katas projects can be disabled with the ENABLE_KATAS 
-if ($Env:ENABLE_KATAS -ne "false") {
-    Get-ChildItem (Join-Path $PSScriptRoot '..') -Recurse -Include '*.sln' -Exclude 'Microsoft.Quantum.Katas.sln' `
-        | ForEach-Object { Build-One $_.FullName }
-} else {
-    Write-Host "##vso[task.logissue type=warning;]Skipping building Katas solutions. Env:ENABLE_KATAS is '$Env:ENABLE_KATAS'."
-}
-
 if (-not $all_ok) {
     throw "At least one test failed execution. Check the logs."
```
tcNickolas, line 29:
```suggestion
    throw "Microsoft.Quantum.Katas project failed the build. Check the logs."
```

scripts/test.ps1
anpaz-msft: Included the DumpUnitary sln to build/tests.
For good measure, I've also added the Katas.sln to the list of tests, even though there are no unittests right now. Just to be consistent that all we build is also targeted for testing.

anpaz-msft:
I'm pretty sure it will save us about 10 minutes, or more generally speaking, the time it takes to build the C# projects.
The overall time will be whatever validation takes longer. [In my last run](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=12260&view=results) that was validating the notebooks on Mac, which took 26 minutes. I've seen that happening in 15 minutes so I suspect we can see build times going down to about 20 mins in avg.

PR #414, opened 2020-07-17
automate running on local server

anirudhakulkarni:
run this python script to start jupyter notebook on local server


anirudhakulkarni:
Thanks fro reviewing!
I will make the suggested changes as earliest as possible.

anirudhakulkarni:
That was a very efficient way of doing it. I should have thought about the same. I have updated the README.md file with those instructions. Thanks for suggestions.

PR #422, opened 2020-07-24
Parallel run of validate notebook

vivanwin:
Speedup is ~10 min for full pipeline run


scripts/validate-notebooks.ps1
```diff
@@ -23,7 +23,9 @@
 [CmdletBinding()]
 Param(
     [Parameter(Position=1)]
-    $Notebook = ""
+    $Notebook = "",
+    [double]$StartIndex = -1.0,
+    [double]$EndIndex = -1.0
```
tcNickolas, line 28: Could you please document the new parameters, similar to the `.PARAMETER Notebook` documentation on lines 17-19? This will help us remember what they do a year from now :-)

Issue #190, opened 2019-10-07
Update the katas to use most recent Q# features


tcNickolas:
Q# evolution introduced some new language features in the past half a year:

* [within... apply... statement](https://docs.microsoft.com/quantum/user-guide/using-qsharp/operations-functions#conjugations)
* [ranges in array slicing](https://docs.microsoft.com/quantum/language/expressions#array-slices)
* features introduced in Q# 0.6 listed in [this blog post](https://devblogs.microsoft.com/qsharp/qsharp-06-language-features-and-more/)

It makes sense to go through the code and check for places where using the new features would improve code readability. Here is the list of all the katas, with ones that have already gone through the update marked as completed.

- [x] Deutsch–Jozsa algorithm (Notebook Tutorial)
- [x] Exploring Grover's search algorithm (Notebook Tutorial)
- [x] Basic quantum computing gates
- [x] Superposition
- [x] Measurements
- [x] Joint measurements
- [x] Teleportation
- [x] Superdense coding
- [x] Deutsch–Jozsa algorithm
- [x] Simon's algorithm
- [x] Grover's algorithm
- [x] Solving SAT problems using Grover's algorithm
- [x] Solving graph coloring problems using Grover's algorithm
- [x] CHSH game
- [x] GHZ game
- [x] Mermin-Peres magic square game
- [x] BB84 protocol
- [x] Phase estimation
- [x] Bit-flip error correcting code
- [x] Truth tables
- [x] Ripple-carry adder
- [x] Unitary Patterns
- [x] QFT
- [x] Distinguishing Unitaries

avinashbhat:
Can I take this up?

tcNickolas:
Yes, absolutely!

It probably makes sense to make a list of katas and go through them one by one, opening separate PRs for each kata that needs updates - this way each step has a smaller scope (and the overall work doesn't feel overwhelming) and the conflicts are less likely to occur.

tcNickolas:
@avinashbhat Could you let us know much progress you've made on this issue? We're hosting a Hackathon tomorrow, and this issue is one of the easier ones to get started with, so if possible, I'd like to split the work on it among several people. On the other hand, I really don't want several people working on the same kata, so splitting it per kata seems like a good idea.

avinashbhat:
It sounds like a good idea! These are the list of katas that are to be updated. I've marked the one I'm working on. It would be great if you reference the PRs back to this comment to track things.
[...]

PR #424, opened 2020-07-27
[DeutschJoszaAlgorithm] Removed unnecessary resets and updated to use…

arkabhat:
… recent Q# features. Updated to most recent Q# features and removed resets as per #190 .


DeutschJozsaAlgorithm/ReferenceImplementation.qs
```diff
@@ -329,8 +323,8 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
             }
             
             // before releasing the qubits make sure they are all in |0⟩ state
-            ResetAll(x);
-            Reset(y);
+            // ResetAll(x);
```
tcNickolas, line 326: One more comment: you can remove these Resets altogether (and the comment before them as well), since the don't make for helpful comments.

PR #426, opened 2020-07-28
[JointMeasurements] Kata is updated with latest functions and Style guidance.

devikamehra:


JointMeasurements/ReferenceImplementation.qs
```diff
@@ -49,49 +50,47 @@ namespace Quantum.Kata.JointMeasurements {
...
+        if (MeasureAllZ(qs) == One){
+            X(qs[1]);  
+		}
     }
     
     
     // Task 7**. Controlled X gate with arbitrary target
     operation ControlledX_General_Reference (qs : Qubit[]) : Unit {
         
-        body (...) {
```
tcNickolas, line 68: This is unfortunately one change we cannot do :-( This operation implements a unitary transformation, and has to have an adjoint variant defined. But at the same time it uses a measurement in it, so Q# compiler cannot compute its adjoint automatically (in general using a measurement renders an operation non-unitary, so it's reasonable behavior). So we need to use this syntax to define the body of the operation and its adjoint separately, rather than use `is Adj` to invoke automatic generation.

I undid this change, so now the CI build should pass.

By the way, thank you so much for trying this! I looked at the CI build failure and noticed that it fails in a different manner than I expected (both project build and notebook validation should fail, and only the latter does, the former passes), and thus found a recently introduced bug in our CI validation - which is great!

devikamehra: This really sounds interesting. Great learning.
Feels good to be of help. 👍

PR #427, opened 2020-07-28
Quantum game notebook expansions

rslstein:


tcNickolas:
Also, it looks like there is some kind of validation error - if it persists after you move the code to the notebook, you should be able to reproduce it locally using https://github.com/microsoft/QuantumKatas/blob/master/scripts/validate-notebooks.ps1 script, then running Check.ipynb notebook it will generate and looking at which cell fails. The CI report doesn't have the actual validation error propagate to the log.

tcNickolas (approved):
This looks great! I did a pass cleaning up the formatting and moving solution descriptions from code comments to Notebook cells (no reason not to take advantage of neat formatting if it's available!). The biggest change was the discussion of GHZ game: the measurements in different basis can confusing, since we don't apply Z gate to perform measurement in Z basis, and we apply H to perform measurement in X basis, so tensor product notations and one of the matrices were off a bit. Other than that, great discussions for both games!

Thank you! I'll merge once the CI passes

PR #429, opened 2020-07-28
[UnitaryPatterns] Updated to use latest Q# features

arkabhat:
Updated some array index slicing to latest Q# syntax.

### DIFF
--- a/UnitaryPatterns/ReferenceImplementation.qs
+++ b/UnitaryPatterns/ReferenceImplementation.qs
@@ -115,17 +115,14 @@ namespace Quantum.Kata.UnitaryPatterns {
     }
     
     // Helper operation: antidiagonal 
-    operation Reflect (qs : Qubit[]) : Unit
-    is Ctl {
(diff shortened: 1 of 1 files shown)


arkabhat:
Will do! I did attempt to change a few of the X-controlled gate-X patterns but they ended up breaking validation and hence I figured I should wait for additional guidance. I will iterate on this, thank you!

arkabhat:
For the Controlled Reflect on lines 128-130, it seems that the Reflect function is not adjoint and hence is not compatible with the ControlledOnInt(0,gate) as a gate. I considered adding adjoint as part of the operation definition, but I realized the internal function does not have an adjoint and hence cannot be autogenerated. Should I create a specific adjoint version of reflect for this purpose?

tcNickolas:
You can modify Reflect signature to be `is Adj+Ctl`, and then replace ApplyToEachC with ApplyToEachCA so that adjoint can be generated automatically

PR #434, opened 2020-07-29
[CI] Fix false positives in projects verification

tcNickolas:
#410 split project build into separate jobs, but missed moving $all_ok variable check from the old script to the new one. This caused build jobs to always report success, even if `dotnet build` fails. This change adds $all_ok variable to store and report build status.


tcNickolas:
For the state before this change, see [this CI build](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=12437&view=results): project does not compile but the error is not propagated to CI build job status.

PR #435, opened 2020-07-29
[PhaseEstimation] Updated Reset statements and style guidance.

devikamehra:


devikamehra:
Update the  Task2.2 Reference implementation.

PR #437, opened 2020-07-29
DJ Algo swapped with BV Algo, 1 comment removed

melmatt7:


DeutschJozsaAlgorithm/DeutschJozsaAlgorithm.ipynb
```diff
@@ -1,8 +1,10 @@
 {
  "cells": [
   {
-   "cell_type": "markdown",
+   "cell_type": "code",
```
tcNickolas, line 4: Looks like all Markdown cells in the notebook somehow got converted into code cells (which, I expect, is causing the errors in CI build, since it's trying to run all code cells). Could you please revert this part of the change?

DeutschJozsaAlgorithm/Tasks.qs
```diff
@@ -162,14 +162,58 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
...
+        // ...
+    }
+    
     //////////////////////////////////////////////////////////////////
-    // Part II. Bernstein-Vazirani Algorithm
+    // Part III. Bernstein-Vazirani Algorithm
     //////////////////////////////////////////////////////////////////
     
-    // Task 2.1. State preparation for Bernstein-Vazirani algorithm
+    // Task 3.1. State preparation for Bernstein-Vazirani algorithm
```
tcNickolas, line 216: I think it's better to keep this task in part II - state prep for both algorithms is the same, so we need to teach it before the first time learner needs it. So this should stay 2.1, and DJ algorithm and running it should be 2.2 and 2.3, respectively.

DeutschJozsaAlgorithm/Tasks.qs
```diff
@@ -162,14 +162,58 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
...
+
+        return isConstantFunction;
     }
     
     
+    // Task 2.2. Testing Deutsch-Jozsa algorithm
+    // Goal: use your implementation of Deutsch-Jozsa algorithm from task 3.1 to test
+    // each of the oracles you've implemented in part I for being constant or balanced.
+    operation DJ_Test () : Unit {
+        // Hint: you will need to use partial application to test oracles such as Oracle_Kth_Qubit and Oracle_ProductFunction;
```
tcNickolas, line 202: We're going to need to swap hints from testing tasks on DJ and BV algorithms - the detailed hint needs to go into the first task the learner sees, and the second one needs to refer to the first one.

DeutschJozsaAlgorithm/DeutschJozsaAlgorithm.ipynb
```diff
@@ -369,31 +393,34 @@
...
   },
   {
-   "cell_type": "markdown",
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {},
+   "outputs": [],
    "source": [
-    "## Part II. Deutsch-Jozsa Algorithm\n",
+    "## Part II. Bernstein–Vazirani Algorithm\n",
```
tcNickolas, line 401: This is really weird, but I just realized that the task order in the notebook is the opposite from the task order in the project version, and it's been this way since day one. So the changes in the notebook are going to be minimal: DJ remains in part II and BV in part III, we just need to change the test numbers in `%kata` commands to match the renumerated tests. Given that all cells became code cells, it might be easiest to revert the change to the notebook altogether and do the small changes to the old version.

DeutschJozsaAlgorithm/DeutschJozsaAlgorithm.ipynb
```diff
@@ -457,14 +463,25 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "open Microsoft.Quantum.Diagnostics;\n",
+    "// Start by implementing a function AllEqualityFactI \n",
```
tcNickolas, line 466: We can remove the part about defining a function altogether - we now have a library function `AllEqualityFactI` in `Microsoft.Quantum.Diagnostics` namespace, so we can recommend to use that one instead.

PR #438, opened 2020-07-29
Updating  version number and instructions to clarify Ctrl+Enter usage

mkostersitz:
Updated the Q# version number, fixed a series of typos and also improved the Ctrl+Enter instructions. 
Renamed UpdateQDKVersion.ps1 to Update-QDKVersion.ps1 so it matches the comments in the PS1 file


scripts/Update-QDKVersion.ps1
```diff
@@ -18,7 +18,8 @@ param(
         
     .EXAMPLE
     
-        PS> ./Update-QDKVersion.ps1 -Version 0.11.2003.3107
+        PS> ./Update-QDKVersion.ps1 -Version
```
tcNickolas, line 21: Please the version number in the example, without the version this command doesn't work (at least on my machine). You could use the latest version here instead of the old one, `0.12.20072031`

tcNickolas: Please also update the script name in the update instructions at https://github.com/microsoft/QuantumKatas/blob/master/.github/CONTRIBUTING.md to match the new name

PR #443, opened 2020-07-29
Add a discussion for, and visualizations of, arbitrary function usage in Deutsch-Josza

anpfeife:
This should cover (at least a good chunk of) the Deutsch-Josza TODO on #328 

Couldn't decide whether this was best located as an addendum to Part 3, where we just finished talking about the nuances of the algorithm, or in Part 5, 'Further Exploration.' Feel free to move it. Also doesn't include any checked code, purely an exposition using some example oracles and the English/non-math intuition that I developed as I was playing around.

PR #375, opened 2020-06-14
[MultiQubitGates] Added Sections on Bra-Ket Algebra and Multi-Qubit Gates acting on Non-adjacent qubits.

jainvasu631:
Added per discussion here: [#328 comment](https://github.com/microsoft/QuantumKatas/issues/328#issuecomment-643769281)

I have added two sections on Ket-Bra Algebra. The sections added include **Bra-Ket Representation** and **Bra-Ket Decomposition**. They are similar to the corresponding sections in SingleQubitGates.

I have added a section on the Usage of Multi-Qubit gates on Non-Adjacent Qubits. It follows the guidelines prescribed by @tcNickolas 

These changes go towards finishing the 2nd checkbox in the checklist [here](https://github.com/microsoft/QuantumKatas/issues/328#issue-602988038).

Issue #328, opened 2020-04-20
Miscellaneous improvements to existing tutorials


tcNickolas:
Here are some ideas for improvements to the [existing tutorials](https://github.com/microsoft/QuantumKatas/tree/master/tutorials):

- [x] Qubit:
   Update section "Demo: Examining Qubit States in Q#" with the new (visual) format of DumpMachine output
- [x] Multi-qubit gates: 
  * [x] Add a section on ket-bra notation for multi-qubit gates
  * [x] Describe the usage of multi-qubit gates on non-adjacent qubits (both in Dirac notation and in matrix notation, the latter with a worked-out 3-qubit example including swapping the qubits, applying the gate on adjacent qubits and undoing the swap)
- [x] Deutsch-Jozsa algorithm:
  What happens if the function is neither constant nor balanced? How exactly does the wavefunction look like for various functions? (Offer the learner some exercises/experimentation around this)
- [ ] Grover search algorithm:
  * [ ] What happens if there are no solutions?
  * [ ] What happens if solutions make up exactly half of the search space?
  * [ ] What happens if solutions make up more than half of the search space? How to handle this?

vivanwin:
Hi can pickup the first point.

sincak:
What kind of system is the %package Microsoft.Quantum.Katas::0.11.2003.3107 command in your Jupyter notebooks supposed to work on? It doesn't work on Linux or Windows (even after loading #Q SDK)

tcNickolas:
@sincak It's supposed to work on both Linux and Windows (and macOS too, I think). Did you do all the steps of installing Jupyter Notebooks and IQ# as described in the [installation guide](https://docs.microsoft.com/en-us/quantum/install-guide/qjupyter), so that running `dotnet iqsharp --version` from command line gives you an IQ# version that matches the one in the `%package` command?

ed1d1a8d:
A couple suggestions for the reference solutions to the PhaseEstimation kata:

- For task 1.4, perhaps a reference solution based on calling `QFT`/`QFTLE` could be given. Currently the reference solution just calls the `QuantumPhaseEstimation` function from the standard library.
- For task 2.2, the reference solution may be more complicated than necessary. You can first use the circuit in 2.1 with U^2 to determine whether the eigenvalue belongs in {1, -1} or {i, -i}, and then you can use the circuit in 2.1 with U and an appropriate rotation to fully determine the eigenvalue.

tcNickolas:
[...]

PR #449, opened 2020-08-03
[ExploringGroversAlgorithm] Covered Issues mentioned in #328

jainvasu631:
This is the last remaining part of Issue #328. I covered MultiQubitGates in Pull Request #375 , @anpfeife covered ExploringDeutschJoszaAlgorithm in Pull Request #443. Only ExploringGroversAlgorithm tutorial was left. 

I have added some heavy math in order to justify my explanations for last 3 cases. I also wanted to add this math earlier but was unsure of the best way to do so. The terse parts are hidden by default and if the reader is interested they can go through it.

Mostly I have based my explanations on my answer [here](https://quantumcomputing.stackexchange.com/questions/9978/changing-qubits-coefficients-to-trigonometric-functions-in-grover-algorithm/13150#13150) and the question I asked [here](https://quantumcomputing.stackexchange.com/questions/12790/grovers-algorithm-on-a-database-with-more-than-50-matching-entries).


tcNickolas (changes requested):
I like the rigorousness of the math - I think about Grover's search algorithm analysis in terms of visual representation myself, so it's a nice compliment to that approach. However, I'm concerned whether this will be accessible to our learners - the contrast between the hands-on approach in the previous sections of the tutorial and the dense math in this section is pretty stark, and the formulas are hard to read even for somebody who is very familiar with the formalism. I'm worried some readers will just drop the tutorial at this point without reaping the benefits.

I would suggest trying to make the math a bit more accessible without sacrificing the rigorousness:
* I'll add the visual variant of the explanation a bit later, similar to the visuals I used in the [new Microsoft Learn module](https://docs.microsoft.com/en-us/learn/modules/solve-graph-coloring-problems-grovers-search/5-grovers-algorithm). I already have the images, so you don't need to work on them yourself - I'll just wait for this change to be merged to avoid merge conflicts.
* Let's use |good⟩ and |bad⟩ state vectors to denote superposition of all solutions and all non-solutions, rather than Greek symbols - that makes it easier for the reader to keep track of which is which (and matches the notation I'll use in the visuals).
* I would use 𝜃 to denote half of the rotation angle, not the whole angle - this should reduce the number of fractions in the formulas (and again, it will match the visuals).
* I suspect some of the sin−1 should actually be arcsin? I prefer to use explicit trigonometric function name "arcsin", to avoid confusion between it and the fraction "1 divided by sin".
* The reader might not be familiar with the << notation, we can write the same thing in words.
* Similarly, I would not use symbol ∀ - it is easily replaced by words, and it's harder to read (I think I last used it in shorthand for math lectures back in university, which didn't aim for general readability)

Let's start with these changes, and I'll try to think of other ways to make this material more readable while keeping it accurate.

Thank you!

jainvasu631:
I will get on this.  I am bit busy during the weekdays. So I will probably work on this o'er the weekend.

jainvasu631:
I have done the suggested changes.

Personally I always found visual explanations bordering on voodoo maths and preferred the cold mathematical rigor. 
However I do agree that this can be challenging to a novice. While I was fixing the changes, I myself was a bit taken aback at the terseness of the arguments. 
You are right about the lack of usage of the symbols  ∀ and << outside the university and college studies. Since I am currently in uni, I am used to cranking out proofs based on induction and formal logic statements and hence in the habit of using these symbols. However anyone outside of academics would find them quite alien. 

The half angle change does reduce the number of fractions but now every iteration of Grovers increases it proximity by 2 theta instead of theta when we did the maths in terms of half angles.

tcNickolas (approved):
First of all, I'm really sorry it took me so long to get back to this review! Oddly enough, I've never done the math to prove this formula, so it took me quite a bit of uninterrupted focus to go through it - and I haven't had that kind of focus time in quite a while.

This is great work, though! I did some minor edits, catching a couple of typos and making some formulas more readable, as well as rearranging the cells to make them easier to edit.

Thank you so much for doing this! And again, apologies for the delay.

Issue #176, opened 2019-10-01
Improve error messaging for the katas that allow it


tcNickolas:
Several katas could have better error messaging:

- [ ] Superposition : the test harness could log the actual state of the system after prep vs the expected state before asserting that they are the same (the change would add state prep and logging using `DumpMachine` to `AssertEqualOnZeroState`)
- [ ] BasicGates : a similar improvement, but a bit more extra code to write a unified test wrapper (since all tests in it just use `AssertOperationsEqualReferenced`)
- [ ] Measurements : the test harness already reports the number of incorrect classifications; in addition, it can print the states which were misclassified in human-readable format, taking an array of string representations of states as an extra parameter to operations like `DistinguishTwoStates_OneQubit` and logging each mismatch.
- [ ] SolveSATWithGrover and GraphColoring : use logging approach similar to RippleCarryAdder to produce more readable results for testing oracles on basis states (so that the solver knows which input state is processed incorrectly, not just that there is an error).

There are probably other katas that allow similar improvements - please share the ideas in the comments!

sandreenko:
I will work on the first task (Superposition).

Omelianenko:
Working on the fourth task, starting from DeutschJozsaAlgorithm

ulitoo:
I will be working on the Measurements test output improvement

vivanwin:
I will work on the BasicGates, to log the actual state.

jackhyder-ms:
DeutschJozsaAlgorithm can be checked off here

jainvasu631:
@tcNickolas What kind of messaging did you have in mind for SolveSATWithGrover?

tcNickolas:
@jainvasu631 Apologies for the delay replying!

For the oracle-based tasks, there are two ways to verify that they are done correctly:
* One can compare the solution to the reference implementation using [AssertOperationsEqualReferenced](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertoperationsequalreferenced) - this is faster but the result it gives is "incorrect qubit state", which is not particularly helpful.
* Or one can iterate through all possible inputs, convert them to basis states, prepare them, apply the oracle to them and measure the results - if the classical result is incorrect (or if the basis states measured changed as a result of applying the oracle), the implementation is incorrect, but you know what's the basis state you need to debug.

[...]

PR #453, opened 2020-08-05
Devikamehra/improve error join measurement

devikamehra:
#176 Improved error statements in Joint Measurement Kata.


JointMeasurements/Tests.qs
```diff
@@ -22,6 +22,9 @@ namespace Quantum.Kata.JointMeasurements {
     operation DistinguishStates_MultiQubit (Nqubit : Int, Nstate : Int, statePrep : ((Qubit[], Int, Double) => Unit is Adj), testImpl : (Qubit[] => Int), preserveState : Bool) : Unit {
         let nTotal = 100;
         mutable nOk = 0;
+        mutable misclassifications = new String[Nstate];
```
tcNickolas, line 25: The CI build fails because it's claiming that `Nstate` identifier doesn't exist. This is a bit weird, since it definitely does exist... Maybe with the next changes this will go away?

JointMeasurements/Tests.qs
```diff
@@ -20,42 +20,71 @@ namespace Quantum.Kata.JointMeasurements {
...
-                if (ans == state) {
-                    set nOk += 1;
+                if ((ans >= 0) and (ans < nStates)) {
+                    // classification result is a valid state index - check if is it correct
+                    if (ans != state) {
+                        set misclassifications w/= ((state * nStates) + ans) <- (misclassifications[(state * nStates) + ans] + 1);
+                    }
                 }
-                
-                if (preserveState) {
```
tcNickolas, line 44: Please keep the logic of `preserveState` variable - you'll notice that the Measurements kata never requires the learner to preserve the input state while learning which state it was, but JointMeasurements kata asks to do this in multiple tasks

JointMeasurements/Tests.qs
```diff
@@ -20,42 +20,71 @@ namespace Quantum.Kata.JointMeasurements {
...
-                    AssertAllZero(qs);
-                } else {
-                    // we're not checking the state of the qubit after the operation
-                    ResetAll(qs);
+                else {
+                    // classification result is an invalid state index - file it separately
+                    set unknownClassifications w/= state <- (unknownClassifications[state] + 1);  
                 }
+                // if we have a max number of measurements per solution run specified, check that it is not exceeded
+                if (measurementsPerRun > 0) {
```
tcNickolas, line 61: The tasks in JointMeasurements kata do not ask for performing the tasks in a limited number of measurements, so we don't need this logic (and the `measurementsPerRun` parameter)

JointMeasurements/Tests.qs
```diff
@@ -76,19 +105,19 @@ namespace Quantum.Kata.JointMeasurements {
     
     // ------------------------------------------------------
     operation T01_SingleQubitMeasurement_Test () : Unit {
-        DistinguishStates_MultiQubit(2, 2, StatePrep_ParityMeasurement, SingleQubitMeasurement, false);
+        DistinguishStates_MultiQubit(2, 2, StatePrep_ParityMeasurement, SingleQubitMeasurement, 0, ["(1/√2)|00⟩ + (1/√2)|11>", "(1/√2)|01⟩ + (1/√2)|10>"]);
```
tcNickolas, line 108: Notice that the task description promises *a superposition* of these basis states, not "an equal superposition" (specific superpositions can be distinguished and restored really easily). A better way to describe these states in a readable format is `["α|00⟩ + β|11⟩", ""α|01⟩ + β|10⟩"]`

JointMeasurements/Tests.qs
```diff
@@ -76,19 +105,19 @@ namespace Quantum.Kata.JointMeasurements {
...
     }
     
     
     // ------------------------------------------------------
     operation T02_ParityMeasurement_Test () : Unit {
-        DistinguishStates_MultiQubit(2, 2, StatePrep_ParityMeasurement, ParityMeasurement, true);
+        DistinguishStates_MultiQubit(2, 2, StatePrep_ParityMeasurement, ParityMeasurement, 0, ["(1/√2)|00⟩ + (1/√2)|11>", "(1/√2)|01⟩ + (1/√2)|10>"]);
```
tcNickolas, line 114: Same here, `["α|00⟩ + β|11⟩", ""α|01⟩ + β|10⟩"]` will work better

JointMeasurements/Tests.qs
```diff
@@ -76,19 +105,19 @@ namespace Quantum.Kata.JointMeasurements {
...
     }
     
     
     // ------------------------------------------------------
     operation T03_GHZOrGHZWithX_Test () : Unit {
-        DistinguishStates_MultiQubit(4, 2, StatePrep_ParityMeasurement, GHZOrGHZWithX, true);
+        DistinguishStates_MultiQubit(4, 2, StatePrep_ParityMeasurement, GHZOrGHZWithX, 0, ["(1/√2)|0000⟩ + (1/√2)|1111>", "(1/√2)|0011⟩ + (1/√2)|1100>"]);
```
tcNickolas, line 120: Please use α and β here as well

JointMeasurements/Tests.qs
```diff
@@ -148,7 +176,7 @@ namespace Quantum.Kata.JointMeasurements {
     
     
     operation T05_DifferentBasis_Test () : Unit {
-        DistinguishStates_MultiQubit(2, 2, StatePrep_DifferentBasis, DifferentBasis, true);
+        DistinguishStates_MultiQubit(2, 2, StatePrep_DifferentBasis, DifferentBasis, 0, ["𝛼|00⟩+𝛽|01⟩+𝛽|10⟩+𝛼|11⟩", "𝛼|00⟩−𝛽|01⟩+𝛽|10⟩−𝛼|11⟩"]);
```
tcNickolas, line 179: Italic Greek letters look odd if the problem uses "normal" Greek letters... Could you use these instead?
```suggestion
        DistinguishStates_MultiQubit(2, 2, StatePrep_DifferentBasis, DifferentBasis, 0, ["α|00⟩ + β|01⟩ + β|10⟩ + α|11⟩", "α|00⟩ - β|01⟩ + β|10⟩ - α|11⟩"]);
```

tcNickolas (changes requested):
I made a pass and left some comments - I think I found the cause of CI build break, so once you fix those, it should work.

Thank you!

tcNickolas (approved):
Looks good! I reverted part of the change that reintroduced deprecated functions RandomInt and RandomReal - they've been deprecated in the most recent QDK release and removed in ebc0ce0225a99d092f3b590011a45f111ed20be7 - and some whitespace changes - those were all minor and didn't warrant another review iteration.

Thank you!

PR #462, opened 2020-08-18
Use PackageReference in all Katas .csproj files

rmshaffer:
### Purpose
With the upcoming IQ# project references feature (see https://github.com/microsoft/iqsharp/pull/262), the IQ# kernel will automatically load any projects and packages specified in the `.csproj` file alongside the current `.ipynb` notebook file.

Some of the Katas projects have a `ProjectReference` to `utilities/Common/Common.csproj`. But this project is also included in the `Microsoft.Quantum.Katas` package. So when the Katas notebooks try to load the `Microsoft.Quantum.Katas` package, this results in duplicate symbol definitions, and an error is thrown.

### Change Description
This PR proposes to avoid this issue by adding a `PackageReference` to `Microsoft.Quantum.Katas` in all of the `.csproj` files, and removing the `ProjectReference` to `utilities/Common/Common.csproj`, if it exists. (The downside of this change is that it would make the local development process slightly more annoying for changes inside `utilities/Common`. However, it seems that changes there are very infrequent, and so I hope this isn't a big deal.)

### Additional Notes
There will be a major end-user benefit of this change soon. Once the IQ# project references feature is available, and the Katas repo is updated to use that version of the QDK, the `%package Microsoft.Quantum.Katas` command at the top of every notebook can be removed. The package will get automatically loaded during IQ# kernel initialization due to the `PackageReference` in the `.csproj`.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -118,7 +118,7 @@ and to [have PowerShell installed](https://github.com/PowerShell/PowerShell#get-
 
 2. **Running on Binder**  
    The Katas can be run online on [Binder](https://mybinder.org); when you make a potentially breaking change (such as an update to the new QDK version or modifying any package dependencies), you need to make sure that this still works.  
-   You can check this by pushing your changes to a branch on GitHub and navigating to the Binder link used for the Katas (https://mybinder.org/v2/gh/Microsoft/QuantumKatas/master?filepath=index.ipynb) and change account name (`microsoft`) and branch (`master`) in the url to your Github username and branch name, respectively. After that you can navigate to the kata you want to check using the links from index notebook.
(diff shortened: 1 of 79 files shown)


rmshaffer:
Moving this back to Draft status; holding off on this change until after the next QDK release.

rmshaffer:
> I think it would be better to add a reference to the Katas **Project** and not the Katas **Package**. Otherwise, it is odd to have to build & create a package just to pick up changes that are built from the same repository.

I agree, @anpaz-msft, it's not ideal. IQ# currently only supports loading magic commands (such as `%kata`) from packages. Initial support for project references in IQ# is limited to only Q# projects. Essentially this change will involve moving the package reference from the notebooks (`%package` command) into the .csproj file (`PackageReference` element).

It is a good point that we should consider in the future for IQ#: adding support for C# project references to get things like custom magic commands, display encoders, etc. Also adding @cgranade, who had mentioned this at one point as well.

rmshaffer:
Actually closing this PR for now, since the purpose of the change will be slightly different when I raise it again in a week or so. I'll move the changes onto a different branch.

PR #463, opened 2020-08-20
[RippleCarryAdder] Fix parameter description + Add missing <b> tags.

theerfan:
I noticed that the "sum" parameter in the description of Task 1.4 of RippleCarryAdder kata was wrongly named, so I fixed it.
Also, some of the "Need a hint? Click here" sections in some katas missed the `<b>` tag (which had been annoying me for some time :D), so I added them.


msoeken (approved):

msoeken:
@theerfan: Let's address the verbosity in a separate PR to split different topics. Thanks for this PR.

PR #465, opened 2020-08-25
Extend ExecutePreprocessor.startup_timeout to 120

ricardo-espinoza:
Increasing the timeout to prevent a condition during CI validation in which a new package takes longer to load.


tcNickolas:
Could you please provide more details on the problem this change should solve? I don't think I've seen this kind of behavior on public CI builds

ricardo-espinoza:
Hi @tcNickolas , this is a change by @rmshaffer that we were testing as part of a possible solution to an issue we encountered with the Solid pipeline during the release process of v0.12.20082513. We should abandon this, so I'm closing the PR.

PR #469, opened 2020-08-26
Replace project references with package references

rmshaffer:
This PR contains the following changes:

- Adds a package reference to `Microsoft.Quantum.Katas` to the .csproj in each kata folder
- Removes references to Common.csproj from each of these .csproj files, which is redundant with the Microsoft.Quantum.Katas package.


rmshaffer:
> I'm looking at running this branch on Binder, and the first `%kata` cell of the Measurements kata takes over 10 minutes to execute. For comparison, the master branch using the same QDK version executes in a couple of seconds. It looks like [this issue](https://github.com/microsoft/iqsharp/issues/79) is back, or a different one, since the execution time is a lot longer than it used to be with that issue.
> 
> Any ideas on what could be causing this?

I have tried a few times but haven't been able to reproduce the issue on Binder; things seem to be running normally. The changes here shouldn't have any runtime impact at all (it's just a build-time change), so if there is some intermittent issue happening on Binder, I would be very surprised if it's related to the changes in this PR...

PR #478, opened 2020-09-04
Use the new test method for all Kata's

vivanwin:


rmshaffer:
@vivanwin looking at these CI failures, I am wondering if there is a similar package load timing issue happening here where the `Microsoft.Quantum.Xunit` package is not being loaded in time.

You could attempt to add the workaround in `validate-notebooks.ps1` that I've tried in my PR #468:
https://github.com/microsoft/QuantumKatas/blob/1b9bcfa00b25602edd82182f50c16ac712ef6115/scripts/validate-notebooks.ps1#L67-L83

I think you would only need to include `Microsoft.Quantum.Standard` and `Microsoft.Quantum.Xunit` in that list to fix this particular issue.

tcNickolas:
Any theory for why this did not happen in the canary case? Not all katas are failing - is it only happening to the katas that use QuantumSimulator rather than CounterSimulator?

rmshaffer:
> Any theory for why this did not happen in the canary case? Not all katas are failing - is it only happening to the katas that use QuantumSimulator rather than CounterSimulator?

I don't think it's correlated to that. I think it may just be a timing thing, whether the packages finish loading before the `.qs` files are compiled. Having the `%workspace reload` in the notebook is likely another way to avoid this problem (some notebooks don't have it).

@vivanwin, the CI failures now are timeouts, which could be addressed by:

- Removing `Microsoft.Quantum.Katas` and `Microsoft.Quantum.MachineLearning` from the `$env:IQSHARP_AUTO_LOAD_PACKAGES` in your `validate-notebooks.ps1` change. Those packages aren't necessary for your scenario here, and will speed up the runtime.
- Adding an increased `--ExecutePreprocessor.startup_timeout` to the `jupyter nbconvert` calls in the `validate-notebooks.ps1` script. I believe it is correct and expected that the startup time will increase as part of this change. We are essentially frontloading all of the work from the `%package Microsoft.Quantum.Katas` and `%workspace reload` cells, as well as loading an additional package `Microsoft.Quantum.Xunit` which is necessary for the Q# tests to compile.

rmshaffer:
The CI is almost there! Only one notebook is failing now: it appears that the `TruthTables` notebook does actually have a compile-time dependency on `Microsoft.Quantum.Katas`, so perhaps that package does actually need to be in the `$env:IQSHARP_AUTO_LOAD_PACKAGES` list after all.

Thanks for working through this!

tcNickolas:
This is _really_ weird - TruthTables doesn't do anything fancy that other katas don't... We'll need to look closer into that

rmshaffer:
> This is _really_ weird - TruthTables doesn't do anything fancy that other katas don't... We'll need to look closer into that

Probably just another unpredictable timing issue, then. 😞

tcNickolas (changes requested):
This problem is not specific to CI builds, I can reproduce it when running the notebook locally - running the first task cell gives `Invalid test name: T1_ProjectiveTruthTables_Test` error. `%workspace reload` cell gives 

    : error IQS000: Could not compile Roslyn dll from working folder.
    : error CS0246: The type or namespace name 'Xunit' could not be found (are you missing a using directive or an assembly reference?)

which seems to point to `Xunit` not loaded in time - but when running the notebook locally, for which we don't have a workaround (I really don't think we can offer the local users to run some setup scripts for a temporary workaround). 

@rmshaffer Have you tried running this with the latest beta of IQ# (the one we're expecting will fix the issue)? I'm concerned that this might point to another problem, since we don't have a good idea where it is coming from.

rmshaffer:
> @rmshaffer Have you tried running this with the latest beta of IQ# (the one we're expecting will fix the issue)? I'm concerned that this might point to another problem, since we don't have a good idea where it is coming from.

Good question, @tcNickolas. I will test this branch against the latest IQ# beta and report back.

rmshaffer:
> I will test this branch against the latest IQ# beta and report back.

It turns out the problem is fairly obvious: somehow `TruthTables.csproj` is not part of this PR, and did not get the `<IQSharpLoadAutomatically>true</IQSharpLoadAutomatically>` property added like the rest of the `.csproj` files did. Once I make that change to `TruthTables.csproj`, everything seems happy.

tcNickolas:
@rmshaffer I see how that could be an issue :-D Thank you for tracking this down! We shouldn't need the workarounds we did if that's the case, right?

@vivanwin Please double-check that the katas that are not included in the CI build are also updated consistently and execute fine when running the corresponding notebooks locally. The list of excluded katas can be found [here](https://github.com/microsoft/QuantumKatas/blob/master/scripts/validate-notebooks.ps1#L95).

Which katas are excluded from this change? (The PR title says "almost all", so it's good to be explicit about which ones are excluded)

rmshaffer:
> @rmshaffer I see how that could be an issue :-D Thank you for tracking this down! We shouldn't need the workarounds we did if that's the case, right?

We shouldn't need the most recent workaround - i.e., I believe `Microsoft.Quantum.Katas` doesn't need to be in `$env:IQSHARP_AUTO_LOAD_PACKAGES` in `validate-notebooks.ps1`. I believe we do need the others still, until after the next QDK release.

vivanwin:
All the Kata's are now using the new @ test method.

rmshaffer (approved):

rmshaffer:
> All the Kata's are now using the new @ test method.
> For now the CI test are a bit slower then it used to be, because for all the tests these packages are loaded regardless if it is needed:
> Microsoft.Quantum.Standard
> Microsoft.Quantum.Xunit
> Microsoft.Quantum.Katas

Thanks @vivanwin! Yes, it looks like the CI is a few minutes slower (increase from ~27 min to ~32 min), but this should become better after the next QDK release, at which point I will remove the workaround to avoid loading unnecessary packages.

scripts/validate-notebooks.ps1
```diff
@@ -58,16 +58,33 @@ function Validate {
     # Convert %kata to %check_kata
     (Get-Content $Notebook -Raw) | ForEach-Object { $_.replace('%kata', '%check_kata') } | Set-Content $CheckNotebook -NoNewline
 
+    # For build 0.12.20082513, work around an IQ# package load timing issue by forcing all necessary packages to be loaded up-front.
```
tcNickolas, line 61: Do we need to do this check separately for each kata, or can we do it once before we start iterating over individual notebooks?

TruthTables/Tests.qs
```diff
@@ -20,46 +20,53 @@ namespace Quantum.Kata.TruthTables {
         Fact(actual::bits == expected::bits, $"Truth table for {name} is not correct");
     }
 
+    @Test("Microsoft.Quantum.Katas.CounterSimulator")
```
tcNickolas, line 23: TruthTables kata doesn't work when executed locally - it gives me "Invalid test name: T1_ProjectiveTruthTables_Test" which typically indicates missing "%workspace reload" command. 

Alternatively, I think this kata should work with `QuantumSimulator`, it doesn't look like it's using any of the `CounterSimulator` features - can you try that?

tcNickolas (changes requested):
Actually, I tried running several different katas locally, and most of them give me "Invalid test name" error - Superposition, Teleportation, CHSHGame, MagicSquareGame and SolveSATWithGrover, to name a few. The katas that *don't* give me this error are BasicGates (don't know why) and Measurements (because it has `%workspace reload`?), so it shouldn't be the case that my local IQ# setup is completely broken.

Have you tried running the notebooks locally from a new command line window (so that leftover environment variable from the validate-notebooks script doesn't work silently)? Can you reproduce the failures I'm seeing when running a fresh setup? 

Let's be careful about this change - we want to keep things working locally that worked before this change :-)

vivanwin:
Let's wait for the new release of IQSharp kernel, this should resolve all these issues.

tcNickolas:
Sounds good! This is not a user-facing change, so waiting a good option :-)

scripts/validate-notebooks.ps1
```diff
@@ -64,6 +64,23 @@ function Validate {
...
+                "Microsoft.Quantum.Standard",
+                "Microsoft.Quantum.Xunit",
+                "Microsoft.Quantum.Katas"
+            ) -join ","
+            Write-Host ("Set IQSHARP_AUTO_LOAD_PACKAGES: " + $env:IQSHARP_AUTO_LOAD_PACKAGES)
+        }
+    } catch {
+        Write-Host ("Failed to read iqsharp version: " + $_)
+    }
+
```
rmshaffer, line 83: Note: this workaround is no longer necessary and can be removed.

TruthTables/Tests.qs
tcNickolas: Let's handle switching from `CounterSimulator` to `QuantumSimulator` in a separate PR - it's going to be easier to keep this a purely syntax change and to review and troubleshoot meaningful changes with smaller scope

tcNickolas (approved):
Looks good! I'll file some follow up issues to remove `CounterSimulator` when `QuantumSimulator` would be sufficient.

Not sure why CI build is stuck, once it passes we can merge.

Thank you!

PR #480, opened 2020-09-09
Build pipeline triggers

anpaz-msft:
Makes sure the build pipeline is triggered for PRs on main/release/feature branches.
Created new GitHub action to sync QDK HEADs on push to main.


tcNickolas:
1) Could you please resolve the conflict in the pipeline trigger?
2) Could you clarify whether this will affect the Katas CI build, or only the end-to-end QDK build?

anpaz-msft:
It affects only the CI build:
- It makes sure that PRs to `main` and any `feature/*` or `release/*` branch triggers a build for validation. It also removes the trigger when a push is done to `main` (since you can't commit to `main` without a PR this is unnecessary and just consumes build host agents).
We're standardizing across all branches on this behavior.

The other file is a GitHub action to trigger a build that automatically updates the `main` branch on the `qdk` repo to point to `main` on all submodules. This guarantees that `main` is always pointing to latest. Similarly, we're standardizing on this across all qdk repos.

anpaz-msft:
Should also mention. Not planning to have an e2e build for Katas. Katas has no downstream dependencies so there is no point in checking if changes to Katas will break anything downstream.
If you feel we should add one, though, we could easily do that.

tcNickolas (approved):
Looks good, thank you for the clarification!

PR #481, opened 2020-09-10
Global replace of references of master to main

anpaz-msft:


tcNickolas (approved):
Looks like the replace tool replaced Unicode symbols with question marks - I fixed that in one file that was affected. Other than that, looks good. Thank you!

PR #483, opened 2020-09-15
Updated 1.14 to add solution 2

katiescott75:
Added solution 2 to 114 Kata in Superposition workbook 2


tcNickolas (approved):
Looks good! CI build was failing because the test name and the operation name were incorrect in the new cell - they should match the ones used in the main task, since we're using the same operation override/test invocation mechanism as in the main kata. I went ahead and fixed that, and a formatting typo that caught my eye in the previous solution explanation.

Thank you! I'll merge once the CI passes.

PR #494, opened 2020-09-29
Update Katas to QDK version 0.12.20100301

rmshaffer:
This PR updates the QDK version to the newly-released v0.12.20100301.

Note that we should be able to resume progress on PRs #468 and #478 after this PR has been merged.

To do:
- [x] Update to released QDK version.


tcNickolas:
Why the cell execution timeout increase? CI build seemed to pass before that. I had the impression the new release offers performance improvement?

rmshaffer:
> Why the cell execution timeout increase? CI build seemed to pass before that. I had the impression the new release offers performance improvement?

If you look at the earlier CI builds, they had intermittent timeouts (which went away on a re-run), so I am just trying to increase the timeout to see if those intermittent timeouts go away.

Unfortunately I still hit the timeout once for a `%package Microsoft.Quantum.Katas` cell, even with the timeout at 300 seconds. See this failure: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=15348&view=logs&j=5be64426-8cdd-541d-c95d-0b4557f8a274&s=d654deb9-056d-50a2-1717-90c08683d50a&t=5a226f14-71f5-5b48-f0c3-560eb5e56c2e&l=248

I don't yet understand why this is happening. We have not been seeing this in the Katas validation that runs as part of the E2E QDK builds (https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build?definitionId=24&_a=summary).

tcNickolas:
Some transient timeouts are normal, sometimes the build agents just don't feel like it; `%package Microsoft.Quantum.Katas` is not... :-(

rmshaffer:
Quick update: the plan on this one is to first update to a new QDK version that includes the fix from https://github.com/microsoft/iqsharp/pull/331, which should resolve the intermittent timeouts we have been seeing in the CI runs here.

tcNickolas:
Sounds good - I'll close it now so that it's less tempting to merge it :-)

Issue #491, opened 2020-09-28
[GroversAlgorithm] Emphasize global phase in "Conditional phase flip" task


tcNickolas:
### Context
[GroversAlgorithm kata](https://github.com/microsoft/QuantumKatas/blob/main/GroversAlgorithm/) has a task "Conditional phase flip", which asks the learner to "Flip the sign of the state of the register if it is not in the |0...0⟩ state.". However, the reference implementation flips the sign of the state if it is in |0...0⟩ state, i.e., implements that transformation up to a global phase of -1. The hint in the task suggests that these two transformations are equivalent, and the tester checks unitary equality up to a global phase.

This is fine if this implementation is used for Grover's search algorithm itself, since the global phase will remain global and not be observable. However, if this implementation is used as a building block for other algorithms, in particular quantum counting algorithm, which converts the global phase difference into the relative phase, this can cause very interesting bugs (see an extended discussion in [this Quantum Computing SE question](https://quantumcomputing.stackexchange.com/questions/5973/counting-in-q-number-of-solutions/6446#6446)).

### Improvement
I would leave the testing harness as is (since for the purposes of Grover's algorithm both implementations are correct!), but put in a bit of effort into clarifying the difference and where it can come into play; so modifying [the hint](https://github.com/microsoft/QuantumKatas/blob/main/GroversAlgorithm/Tasks.qs#L138) (and in the matching notebook) to explain that for the purpose of Grover's algorithm these transformations will be the same, but if you're using it for something else you'll need an exact implementation. 
I'd also modify [both reference solutions](https://github.com/microsoft/QuantumKatas/blob/master/GroversAlgorithm/ReferenceImplementation.qs#L90) to include a comment along the lines "To fix the global phase difference, use the following line" and a line that uses the [R gate](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.r) to add the right global phase. This way the reference solutions would implement the exact transformation.

Manvi-Agrawal:
Raising a new PR due to git issues

Manvi-Agrawal:
Recreating a new PR due to git issues

PR #505, opened 2020-10-04
Emphasized the importance of global phase in "Conditional phase flip" task in Grover's Algorithm Kata

Manvi-Agrawal:
Pull request in regard to the following issue : #491


GroversAlgorithm/GroversAlgorithm.ipynb
```diff
@@ -306,6 +306,15 @@
     "    You need to modify it to add the $-1$ phase to only the $|0...00\\rangle$ state instead.\n",
     "    <br/><br/>\n",
     "    Alternatively, you can use the same trick as in the oracle converter task.<br>\n",
+    "</details>\n",
+    "<br>\n",
+    "\n",
+    "<details>\n",
+    "  <summary><b>IMPORTANT : The test will pass but its not the correct implementation. For More details click here</b></summary>\n",
```
tcNickolas, line 313: I would move this comment to be part of the 1st note, so that the learner gets the information "let's implement a different transformation" and "beware, it is different and here are the consequences" in the same batch of hints.

tcNickolas: Also, the note "What we've just told you to do is not correct" looks a bit weird :-)

GroversAlgorithm/ReferenceImplementation.qs
```diff
@@ -13,6 +13,7 @@ namespace Quantum.Kata.GroversAlgorithm {
     open Microsoft.Quantum.Arrays;
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Math; // To use PI
```
tcNickolas, line 16: I wouldn't leave a comment for the import - we usually import libraries when they are needed without much justification.
```suggestion
    open Microsoft.Quantum.Math;
```

GroversAlgorithm/ReferenceImplementation.qs
```diff
@@ -94,6 +95,10 @@ namespace Quantum.Kata.GroversAlgorithm {
         // Convert it into a phase-flip oracle and apply it
         let flipOracle = OracleConverter_Reference(allZerosOracle);
         flipOracle(register);
+        
+        // To fix the global phase difference, use the following line :
+        // For more details refer to the following Quantum SE question : https://quantumcomputing.stackexchange.com/questions/5973/counting-in-q-number-of-solutions/6446#6446
+        R(PauliI, 2.0 * PI(), register[0]);
```
tcNickolas, line 101: Can you add this line to `PhaseFlip_ControlledZ` as well? I believe that alternative solution suffers from the same issue

tcNickolas (changes requested):
I made a pass and left a couple of comments. Also, could you add the same note to Tasks.qs variant of the kata (https://github.com/microsoft/QuantumKatas/blob/main/GroversAlgorithm/Tasks.qs#L138)? Some people might use the Q# project version of the kata, so we're trying to keep both "frontends" of each kata in sync.

Thank you!

GroversAlgorithm/GroversAlgorithm.ipynb
Manvi-Agrawal: I have done the changes. Also I added the mathematical part in Hint3. Let me know if any further changes need to be done.

GroversAlgorithm/ReferenceImplementation.qs
Manvi-Agrawal: Done

GroversAlgorithm/ReferenceImplementation.qs
Manvi-Agrawal: Done

PR #506, opened 2020-10-05
Fixed expression in properties of Pauli gates in SingleQubitGates.ipynb

valdaarhun:
Fixed expression in anti-commutative property of Pauli gates.


vivanwin (approved):
Great catch thank you.

PR #508, opened 2020-10-06
[GraphColoring] Fixed missing right hand bracket ')' in Task 1.4

jainvasu631:
While I was working on the Workbook, I noticed this error. This PR is the fix for that.


tcNickolas (approved):
Nice catch! (Looks like our CI builds are not triggered, so I'll skip the CI and merge with local validation only.)

Thank you!

Issue #208, opened 2019-10-28
Create workbooks for the katas/tutorials, part 1


tcNickolas:
The katas offer a set of programming exercises on quantum computing and reference solutions to them, but we often hear that it would be helpful to have the solutions explained, with the logic steps necessary to arrive from the problem description to the code spelled out. It is especially important for people who go through the katas on their own, without a study group to support them.

To address this, we came up with the idea of "workbooks" - explanations of the solutions to the tasks offered in the katas/tutorials, written in Jupyter Notebook format. You can see the workbook for Superposition kata [here](https://github.com/microsoft/QuantumKatas/blob/master/Superposition/Workbook_Superposition.ipynb) - it is still a work in progress, but it already shows the idea of the workbook, the kind of the content it should have and the way it accompanies the kata.

There are several other katas/tutorials that are a great fit for this kind of learning material:

- [x] Complex arithmetic (classical)
- [x] Linear algebra (classical)
- [x] Single-qubit gates
- [x] Multi-qubit systems
- [x] Multi-qubit gates
- [x] Superposition
- [x] Measurements
- [x] Joint measurements
- [x] SolveSATWithGrover
- [x] GraphColoring
- [x] CHSH game
- [x] GHZ game
- [x] Unitary patterns

Other katas might fit this scheme a bit worse - for example, Superdense Coding kata follows the structure of superdense coding algorithm, so a tutorial on the end-to-end algorithm would work better than independent explanations of individual tasks.

vivanwin:
If there is still help needed I could start working on the first two.

tcNickolas:
Nobody is working on those yet, help will be most definitely appreciated!

eloiseespel:
I would love to help too, I'll start working on Measurements!

tcNickolas:
Sounds great!

@vivanwin and @eloiseespel Please don't feel obliged to do the whole workbook at once (it's a lot of work!) - you can do the first couple of tasks and send them for review to get early feedback. Thank you again for volunteering!

tcNickolas:
@eloiseespel How are you doing with the Measurements workbook? Do you need any help? Please don't hesitate to reach out with questions! And if you ran out of time after doing just a few tasks, feel free to send them in - Measurements is a massive batch of problems, and doing it in one go is an ambitious endeavor!

eloiseespel:
[...]

PR #509, opened 2020-10-06
[GraphColoring] Workbook Part1 (Tasks 1.1 - 1.5)

jainvasu631:
This PR is as per the discussion [here](https://github.com/microsoft/QuantumKatas/issues/208#issuecomment-703868896)
It includes solutions and explanations for Tasks 1.1 - 1.5
I have written a different solution for Task 1.4 as compared to ReferenceImplementation in order to create a flow for the next task 1.5

It creates a partial Workbook for GraphColoring Kata with @BrokenDuck working on the other half consisting of Tasks 2.1-2.3.


tcNickolas:
I'm doing a cleanup pass on the workbook now, and discovered a bug which causes the reference solution for task 1.5 to fail. I sent a PR to fix it (#527), once that one is merged the CI failure should be fixed.

tcNickolas (approved):
Looks good. I did a round of cleanup, using the workbook-style introduction in the workbook (rather than kata-style introduction) and polishing the explanations a bit. This should be good to go once #527 is merged.

Thank you!

PR #515, opened 2020-10-07
Improve CI performance

rmshaffer:
This change improves the performance of Katas CI runs by caching NuGet packages in advance via `dotnet restore` and then clearing all NuGet package sources while executing the notebooks.

This should reduce the total CI runtime to around 15 minutes in the typical case.


tcNickolas (approved):
Looks good, and I love the 3x CI build speedup - very timely for all the Hacktoberfest contributions coming in! Thank you!

PR #516, opened 2020-10-07
Ensure that all required packages are cached for Binder builds

### LINKED ISSUE #511: QuantumClassification tutorials don't work when using Docker image (e.g., Binder)
Example: Open the "InsideQuantumClassifiers" tutorial on Binder:
https://mybinder.org/v2/gh/Microsoft/QuantumKatas/main?filepath=tutorials/QuantumClassification/InsideQuantumClassifiers.ipynb

When executing the third cell, the following error appears:
```
/snippet_.qs(4,6): error QS6104: No namespace with the name "Microsoft.Quantum.MachineLearning" exists.
/snippet_.qs(15,29): error QS5022: No identifier with the name "InputEncoder" exists.
```

I believe this is because the [`Dockerfile`](https://github.com/microsoft/QuantumKatas/blob/main/Dockerfile) currently doesn't pre-build any Katas that depend on `Microsoft.Quantum.MachineLearning`, which means that package does not get downloaded and cached in the image. Since the `Dockerfile` then disables all NuGet package downloading, the package cannot be retrieved at runtime.

I expect the fix for this is just to add a command to the `Dockerfile` something like:
```
./scripts/prebuild-kata.sh tutorials/QuantumClassification InsideQuantumClassifiers.ipynb
```

rmshaffer:
For the Docker build (used by Binder), first run `dotnet restore` for each project. This ensures that necessary packages are cached for *all* Katas, even the ones that aren't prebuilt.

This resolves #511 because `dotnet restore` is called for the `tutorials/QuantumClassification` project as part of this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN pip install -I --no-cache-dir \
 USER ${USER}
 
 RUN cd ${HOME} && \
-# `dotnet restore` for each folder to ensure NuGet cache is fully populated
+# `dotnet restore` for each solution to ensure NuGet cache is fully populated
     for solution in $(find . -type f -name "*.sln"); do dotnet restore "$solution"; done && \
 # Pre-exec notebooks to improve first-use start time
 # (the katas that are less frequently used on Binder are excluded to improve overall Binder build time)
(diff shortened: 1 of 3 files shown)


tcNickolas (approved):
Thank you!

PR #517, opened 2020-10-07
Addition of Graph Coloring Workbook

BrokenDuck:
Finished Graph Coloring Workbook, with comprehensive solutions for every task and done in the standard format.
Part 1 done by : @jainvasu631 
Part 2 done by : me


tcNickolas:
This PR is going to conflict with the final version of the first part of the workbook added in #509. Could you please take that final version and add your solutions to it, so that the pull request shows only the additions? Thank you!

PR #523, opened 2020-10-09
#492 [CHSHGame/GHZGame] Repeat randomized tests multiple times

### LINKED ISSUE #492: [CHSHGame/GHZGame] Repeat randomized tests multiple times
Tasks 2.2 and 2.4 in [CHSHGame kata](https://github.com/microsoft/QuantumKatas/blob/main/CHSHGame/Tests.qs) and 2.2 in [GHZGame kata](https://github.com/microsoft/QuantumKatas/blob/main/GHZGame/Tests.qs) can have incorrect solutions (for example, measuring in incorrect basis) that still could pass the tests with non-negligible probability. It would make sense to repeat the test (the operations inside the `using` block) several times to reduce the probability of the tests giving false positive verdict.

ziqi-ma:
Fixes #492

### DIFF
--- a/CHSHGame/Tests.qs
+++ b/CHSHGame/Tests.qs
@@ -74,10 +74,27 @@ namespace Quantum.Kata.CHSHGame {
     @Test("QuantumSimulator")
     operation T22_AliceQuantum_Test () : Unit {
         using (q = Qubit()) {
+            // repeat 4 times since we are testing a measurement and wrong basis still might get
+            // the correct answer, this cuts the error probability to ~6%
+            EqualityFactB(AliceQuantum(false, q), false, "|0⟩ not measured as false");
+            Reset(q);
+            EqualityFactB(AliceQuantum(false, q), false, "|0⟩ not measured as false");
+            Reset(q);
+            EqualityFactB(AliceQuantum(false, q), false, "|0⟩ not measured as false");
+            Reset(q);
             EqualityFactB(AliceQuantum(false, q), false, "|0⟩ not measured as false");
             Reset(q);
 
             // apply the Pauli X gate
+            X(q);
+            EqualityFactB(AliceQuantum(false, q), true, "|1⟩ not measured as true");
+            Reset(q);
+            X(q);
+            EqualityFactB(AliceQuantum(false, q), true, "|1⟩ not measured as true");
+            Reset(q);
+            X(q);
+            EqualityFactB(AliceQuantum(false, q), true, "|1⟩ not measured as true");
+            Reset(q);
             X(q);
             EqualityFactB(AliceQuantum(false, q), true, "|1⟩ not measured as true");
             Reset(q);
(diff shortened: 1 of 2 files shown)


tcNickolas (changes requested):
Can you please use a for loop to repeat the whole block inside `using` 4 times, rather than repeating each instruction 4 times verbatim? You can use an anonymous variable for the loop variable, like this: `for (_ in 1 .. 4)`

Thank you!

ziqi-ma:
> Can you please use a for loop to repeat the whole block inside `using` 4 times, rather than repeating each instruction 4 times verbatim? You can use an anonymous variable for the loop variable, like this: `for (_ in 1 .. 4)`
> 
> Thank you!

Addressed - also realized the error probability is already ~6% now it should be 6%^4

CHSHGame/Tests.qs
```diff
@@ -117,24 +121,28 @@ namespace Quantum.Kata.CHSHGame {
...
-
-            RotateBobQubit_Reference(true, q);
-            EqualityFactB(BobQuantum(true, q), false, "-π/8 from |0⟩ not measured as false");
-            Reset(q);
-
-            X(q);
-            RotateBobQubit_Reference(true, q);
-            EqualityFactB(BobQuantum(true, q), true, "-π/8 from |1⟩ not measured as true");
-            Reset(q);
+        for _ in (1 .. 4) {
```
tcNickolas, line 124:
```suggestion
        for (_ in 1 .. 4) {
```

CHSHGame/Tests.qs
```diff
@@ -73,25 +73,29 @@ namespace Quantum.Kata.CHSHGame {
...
-            EqualityFactB(AliceQuantum(true, q), false, "|+⟩ is not measured as false");
-            Reset(q);
-
-            // apply the Pauli X and then the Hadamard gate
-            X(q);
-            H(q);
-            EqualityFactB(AliceQuantum(true, q), true, "|-⟩ is not measured as true");
-            Reset(q);
+        for (_ in 1 .. 4) {
+            // repeat 4 times since we are testing a measurement and wrong basis still might get
```
tcNickolas, line 77: I'm not sure what's the best way to estimate the probability of a false positive in one iteration, since the learner can come up with some creative solutions... I'd just say something like "repeat the test several times to reduce the probability of a false positive" without exact numbers.

CHSHGame/Tests.qs
ziqi-ma: hmm if you go to line 139 I'm following that

ziqi-ma: oops addressed

CHSHGame/Tests.qs
ziqi-ma: addressed

tcNickolas (approved):
Looks good - CI build is a transient failure, I'm rerunning it and will merge once it passes.

Thank you!

Issue #522, opened 2020-10-09
Specify the task1 of TruthTable kata clearly in qsharp and jupyter frontend


Manvi-Agrawal:
Currently, the Truth Tables kata asks learner to define two new truth tables x2 and x3, but the intention is not clearly specified. 

Therefore some description can be added, specifying that :<br>
x2 should be true for : 010,011,101,110
x3 should be true for : 100,101,110,111

Manvi-Agrawal:
Would be happy to send in a fix.

msoeken:
I think that this would give away too much information. The task states that `x2` and `x3` are projective functions for elementary variables, and provides the on-set for `x1` as an example. The readers of this task should determine the on-sets should determine the on-sets for `x2` and `x3` themselves. One might point to the `x2` column in the table above as a hint.

Manvi-Agrawal:
Thanks @msoeken for the clarification

Manvi-Agrawal:
Hi, @msoeken @tcNickolas , I was thinking if we could add a couple of new tasks in Truth Tables kata. 
Honestly I was confused about the intension of task1. From the example in the explaination of truth table of  (x1 AND x2), I was expecting a similar kind of task, ie ( (x1 AND x2 AND x3), (x1 AND x2 AND x3) or any descriptive task like either one of x1,x2 and x3 is true). 

Also, I felt example of x1 to be a bit misleading. In theory we were mentioning projective functions, f as a mapping from n-inputs to a single output. So, I misunderstood first task as the 
Given : x1 should be true for : 001,011,101,111
Goal : To specify the truth table for x1 in binary form.

Some of the new tasks which I think can be added 
- Specify truth table for 3-bit input function (x1 AND x2 AND x3)
- Specify truth table for 3-bit input function if exactly 1 bit out of x1, x2 or x3 is true
- Specify truth table for 3-bit input function if exactly 2 bits out of x1, x2 or x3 is true

**Another suggestion**
Also, I think it would be nice if we could explain AND, OR, XOR to the learner by providing the truth tables for 2 bits instead of just asking them to use the bitwise operators. Although, I am not sue whether we wuld want to do this in the Truth Tables kata, since katas are supposed to provide high level implementation. What are your about making a tutorial for the same?

msoeken:
[...]

PR #524, opened 2020-10-09
Improved explaination for the projective function task in Truth Tables Kata

Manvi-Agrawal:
In regard to the following issue https://github.com/microsoft/QuantumKatas/issues/522


TruthTables/Tasks.qs
```diff
@@ -44,19 +44,23 @@ namespace Quantum.Kata.TruthTables {
 
     // Task 1. Projective functions (elementary variables)
     //
-    // Goal: Describe the three projective functions x₁, x₂, x₃ as 3-input
-    //       functions, represented by integers.  Note that we follow the
-    //       convention that x₁ is the least-significant input.
+    // Goal: Describe the three projective functions f₁, f₂, f₃ represented
+    // by integers. They are 3 input functions i.e. f(x) : {0,1}ⁿ → {0,1},
```
tcNickolas, line 48: You can write this more concisely using Unicode (my go-to page for task descriptions is https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts):
```suggestion
    // by integers. They are 3 input functions i.e. f(x) : {0,1}³ → {0,1}.
```

TruthTables/Tasks.qs
```diff
@@ -44,19 +44,23 @@ namespace Quantum.Kata.TruthTables {
...
+    // n being equal to 3.
     //
-    // Example: The function x₁ (least-significant input) is given as an
-    //          example.  The function is true for assignments 001, 011, 101,
-    //          and 111.
+    // f₁ should be true for the following assignments : 001, 011, 101 and 111.
```
tcNickolas, line 51: I think it's a lot less interesting to list the inputs for which the function should be true than to explain what this function is (function f₁ returns the first input, f₂ - the second one, etc.) and let the learner figure out this list themselves. That's why we give one example but not all three.

I would make it more clear what x₂ and x₃ are, but not spell out for which inputs they should be 1.

TruthTables/Tasks.qs
```diff
@@ -44,19 +44,23 @@ namespace Quantum.Kata.TruthTables {
...
+    // f₃ should be true for the following assignments : 100, 101, 110 and 111.
+    //
+    // Note: We follow the convention that x₁ is the least-significant input.
+    //
+    // Example: The function f₁ is given as an example.
     function ProjectiveTruthTables () : (TruthTable, TruthTable, TruthTable) {
-        let x1 = TruthTable(0b10101010, 3);
-        let x2 = TruthTable(0, 0);           // Update the value of x₂ ...
-        let x3 = TruthTable(0, 0);           // Update the value of x₃ ...
+        let f1 = TruthTable(0b10101010, 3);
```
tcNickolas, line 59: I believe it is customary to denote the function that returns the least significant bit of the input as x1 rather than f1. @msoeken should be able to confirm.

tcNickolas (changes requested):
I am happy to have the intention of the task clarified, but I don't think we should spell it out completely - the idea is for the learner to get practice converting multi-bit inputs to indices of bits in the number that represents the function. 

@msoeken Could you take a quick look at this PR to make sure we preserve the intent and notations of your design?

TruthTables/Tasks.qs
msoeken: Yes, I confirm, because these truth tables represent the elementary functions for input variables.

TruthTables/Tasks.qs
```diff
@@ -44,19 +44,23 @@ namespace Quantum.Kata.TruthTables {
...
+    // Goal: Describe the three projective functions f₁, f₂, f₃ represented
+    // by integers. They are 3 input functions i.e. f(x) : {0,1}ⁿ → {0,1},
+    // n being equal to 3.
     //
-    // Example: The function x₁ (least-significant input) is given as an
-    //          example.  The function is true for assignments 001, 011, 101,
-    //          and 111.
+    // f₁ should be true for the following assignments : 001, 011, 101 and 111.
+    // f₂ should be true for the following assignments : 010, 011, 110 and 111.
+    // f₃ should be true for the following assignments : 100, 101, 110 and 111.
```
msoeken, line 53: This gives away too much information (see also my comment in https://github.com/microsoft/QuantumKatas/issues/522#issuecomment-706050763)

msoeken (changes requested):
This gives a way too much information to solve the task. Variable names should be `x_i` instead of `f_i`. Providing some more information can be interesting though.

TruthTables/Tasks.qs
msoeken: Agree. I didn't read the comments in order and gave a similar comment below.

Manvi-Agrawal:
Hi @tcNickolas @msoeken made changes in the qsharp front-end as per the review comments. Currently making changes in the jupyter notebook. Could you pleas go through the edits and suggest any changes.

Manvi-Agrawal:
Attaching screenshots of the hints for better understanding of the diffs in jupyter notebooks :

Screenshot of Truth Table for Hint 1
Screenshot (209)

Screenshot for Truth Table for Hint 4
Screenshot (211)

tcNickolas (approved):
Looks good - this way it reveals less of a solution while helping the learner to arrive to it. But it looks like this kata is indeed ripe for a workbook... I did the usual bit of cleanup, for example, adjusted the tables so that similar columns would have same width.

Thank you!

Issue #485, opened 2020-09-24
Update notebooks validation script to exclude individual tasks


tcNickolas:
### Context
Currently [validate-notebooks.ps1](https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1) script excludes several notebooks which would fail validation. There are several reasons for failure (see them documented [here](https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1#L84)), and often they affect one or two tasks from the notebook, rather than all of them. Excluding whole notebooks from validation also means that CI can miss issues like invalid JSON in the notebook itself (see an example in #411).

### Improvement
Modify the script so that it's possible to exclude from validation individual tasks, rather than whole notebooks. This would allow us to improve the test coverage by including tasks that can be validated and validating the notebooks formatting. For example, in CHSHGame kata only task 1.2 cannot be validated (since it consists of two cells that would need to be implemented simultaneously, which is not possible with the current implementation), the rest of the tasks can be validated normally.

### Implementation
* The tasks are validated by using `%check_kata` magic in each cell that uses `%kata`, so we shouldn't leave either of these magics in the cell before its validation. 
* We could either remove the magic line altogether for the specified tasks (so that the remaining cell would only define an operation, not run the test on it - which can be an issue if the code skeleton in the cell is invalid), or remove the whole cell for those tasks. This requires some experimentation.
* The syntax of defining the tasks in the script could be something like `'CHSHGame.ipynb#T12_ClassicalStrategy_Test'` instead of the current `'CHSHGame.ipynb'`.

Manvi-Agrawal:
Hi @tcNickolas. Is it necessary to specify the tasks that need to be excluded in the <code>validate_notebooks.ps1</code> script as <code>CHSHGame.ipynb#T12_ClassicalStrategy_Test</code>? <br>

I have a slightly different strategy to solve the problem. 
- Firstly add a tag to the <code>exclude_from_validation</code> to the operations <code>AliceClassical</code> and <code>BobClassical</code> in <code>CHSHGame.ipynb</code> as shown below : 
![excludeFromValidation](https://user-images.githubusercontent.com/40084144/96266529-5e100380-0fe4-11eb-9f96-b671476dae0b.png)

- Then, update the <code>validate_notebooks.ps1</code> script as follows :<br>
<pre>
<code>
if ($env:SYSTEM_DEBUG -eq "true") {
[...]

PR #527, opened 2020-10-10
[GraphColoring] Fix qubit allocation check in task 1.5

tcNickolas:
We need to check that the solution doesn't use more than 2N+1 qubits passed to it as an argument.
This check has to happen before AssertOperationsEqualReferenced call, since that operation allocates twice that many qubits to use Choi–Jamiłkowski isomorphism.

This was introduced in #446, and somehow I missed that the reference solution fails the test. GraphColoring kata is excluded from CI validation. This emphasizes the need for fixing https://github.com/microsoft/QuantumKatas/issues/485.


vivanwin (approved):
Good catch, thank you!

Issue #507, opened 2020-10-05
[SolveSATWithGrover] Refactor tests for 1.6 and 2.2 to run faster


tcNickolas:
Currently the tests for tasks [1.6](https://github.com/microsoft/QuantumKatas/blob/main/SolveSATWithGrover/Tests.qs#L264) and [2.2](https://github.com/microsoft/QuantumKatas/blob/main/SolveSATWithGrover/Tests.qs#L330) generate the SAT instances randomly, with up to 7 variables and 14 clauses, which can cause significant test slowdown if several tests are generated on the larger side (the more variables and clauses the test has, the longer it runs). 

It makes sense to refactor the tests so that they run faster without loss of coverage.

1. Task 1.6 doesn't really need to spend that much time on 2-SAT instances (that's an artifact left from before when task 1.5 was dealing not with an individual clause but with 2-SAT problem instead), so we can replace `Run2SATTests` with `RunCrossTests` which only include OR, XOR and AlternatingBits oracles, and drop the random 2-SAT test instances.
2. Both tasks 1.6 and 2.2 should generate the SAT instances more deterministically (though we don't need them to be 100% deterministic). I would modify `Generate_SAT_Instance` to take two parameters - the number of variables and the number of clauses - and pass them manually from the test so that we can have a fixed distribution of tests in each run. I would also reduce the number of tests - maybe one random test for each of 3 .. 6 variables for task 1.6 and one random test for each of 2 .. 6 variables for task 2.2. (I would also rename `Generate_SAT_Instance` to `GenerateSATInstance`, since here the underscores don't carry any particular meaning.)

sgalpha01:
Hey @tcNickolas, can you please assign this issue to me? It would be my good first issue in this topic. Also, I may need a bit of help as I am a beginner.

PR #528, opened 2020-10-10
Refactoring SolveSATWithGrover tests

ziqi-ma:
Fixing #507


SolveSATWithGrover/Tests.qs
```diff
@@ -249,35 +248,23 @@ namespace Quantum.Kata.GroversAlgorithm {
...
+        RunCrossTests(Oracle_SAT);
 
-        // General SAT instances
-        for (i in 1..5) {
-            let (nVar, problem) = Generate_SAT_Instance(-1);
-            Message($"Testing k-SAT instance ({nVar}, {SATInstanceAsString(problem)})...");
+        // General SAT instances for 3..6 variables
 
-            AssertOracleImplementsFunction(nVar, Oracle_SAT(_, _, problem), F_SAT(_, problem));
+        for (i in 3..6) {
```
tcNickolas, line 261: I would call the loop variable `nVar` here - since we switched to iterating over the number of variables in the clause rather than just test index, we can use a more meaningful name. (This will also reduce the scope of changes in the later lines)
```suggestion
        for (nVar in 3..6) {
```

SolveSATWithGrover/Tests.qs
```diff
@@ -249,35 +248,23 @@ namespace Quantum.Kata.GroversAlgorithm {
...
+            let problem = GenerateSatInstance(i,3,-1);
```
tcNickolas, line 262: The operation name mismatch was causing the build failure
```suggestion
            let problem = GenerateSATInstance(i, 3, -1);
```

tcNickolas: I'd also add some variance in the number of clauses - it's better to cover different scenarios
```suggestion
            let problem = GenerateSATInstance(nVar, nVar - 1, -1);
```

SolveSATWithGrover/Tests.qs
```diff
@@ -336,13 +323,13 @@ namespace Quantum.Kata.GroversAlgorithm {
     @Test("QuantumSimulator")
     operation T22_Oracle_Exactly1SAT_Test () : Unit {
         // General SAT instances
-        for (i in 1..10) {
-            let (nVar, problem) = Generate_SAT_Instance(3);
-            Message($"Testing exactly-1 3-SAT instance ({nVar}, {SATInstanceAsString(problem)})...");
+        for (i in 2..6) { // test for 2..6 variables
```
tcNickolas, line 326: I would move the "for 2 .. 6 variables" part of the comment to the previous line, so that there is no inline comment (and it looks more like task 1.6)

SolveSATWithGrover/Tests.qs
```diff
@@ -336,13 +323,13 @@ namespace Quantum.Kata.GroversAlgorithm {
...
+            let problem = GenerateSATInstance(i,3,3); // 3 clauses, and each clause 3 terms
```
tcNickolas, line 327: Same here, I'd suggest using `nVar` for loop variable and some variance in the number of clauses
```suggestion
            let problem = GenerateSATInstance(nVar, nVar - 1,3);
```

tcNickolas (changes requested):
Looks good - I did a pass and left some comments to fix the build break and make the code a bit easier to read.

Thank you!

SolveSATWithGrover/Tests.qs
ziqi-ma: Addressed in latest commit

SolveSATWithGrover/Tests.qs
ziqi-ma: Addressed in latest commit

SolveSATWithGrover/Tests.qs
ziqi-ma: Addressed in latest commit

tcNickolas (approved):
Looks better! There was one more issue - you were passing `-1` as the number of clauses rather than `nVar - 1`, so the code compiled but the tests were failing when they tried to generate a SAT problem with -1 clauses. I fixed that - seemed not worth another round of edits and review :-)

Thank you!

ziqi-ma:
> Looks better! There was one more issue - you were passing `-1` as the number of clauses rather than `nVar - 1`, so the code compiled but the tests were failing when they tried to generate a SAT problem with -1 clauses. I fixed that - seemed not worth another round of edits and review :-)
> 
> Thank you!

Oh somehow I thought passing -1 makes it generate a random number...

Issue #488, opened 2020-09-26
Update notebooks validation script to validate workbooks solutions


tcNickolas:
### Context
Currently [validate-notebooks.ps1](https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1) script validates the workbook code cells using the same logic as the code cells in notebook versions of the katas: `%kata` magic that invokes test for the code in the cell is replaced with `%check_kata` magic that invokes test for the reference solution (from `ReferenceImplementation.qs`). This means that validating the workbooks ignores the solutions in them and does double-coverage of reference solutions. 

### Improvement
Modify the script to validate the actual solutions in the workbooks.

### Implementation
I believe it should suffice to not replace `%kata` with `%check_kata` for notebooks that have `Workbook` as part of their name at https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1#L59. This requires careful testing, though, including adding deliberately incorrect code to the workbooks and making sure this fails the CI build.

Manvi-Agrawal:
Hi @tcNickolas. 

I would prefer this issue to get resolved before the issue https://github.com/microsoft/QuantumKatas/issues/485 as it effects the 2 workbooks(CHSHGame and SolveSATWithGrover)

I have one approach in mind :

- Copy the workbooks to Check.ipynb validation artifact...
<pre>
<code>
if($Notebook -match 'Workbook_*') {
		# Write workbook data to Check.ipynb artifact
		(Get-Content $Notebook -Raw) | Set-Content $CheckNotebook -NoNewline
	}
	else{
		# Convert %kata to %check_kata
		(Get-Content $Notebook -Raw) | ForEach-Object { $_.replace('%kata', '%check_kata') } | Set-Content $CheckNotebook -NoNewline
	}
</code>
</pre>

tcNickolas:
Sounds good!

Manvi-Agrawal:
Cool will raise a PR shortly...

PR #529, opened 2020-10-11
Add graph coloring part 2

BrokenDuck:
Part 1 done by @jainvasu631


tcNickolas (approved):
Looks good - I did a round of cleanup fixing some typos, improving the explanations flow and making sure the reference solutions work (https://github.com/microsoft/QuantumKatas/issues/488 means that CI doesn't check the code in the workbooks at all, so it has to be validated by hand.) I also added the new workbook to the list of exceptions in the validate-notebooks.ps1 script, since task 2.3 would time out quite often in CI.

Thank you!

PR #539, opened 2020-10-18
Updating notebooks validation script to exclude individual tasks

Manvi-Agrawal:
Currently <code>validate-notebooks.ps1</code> script excludes several notebooks in which some tasks would fail validation due to several reasons : 
- Some tasks require implementing two code cells at once before running the test, so the first of the cells implemented is guaranteed to fail
-  There are some tasks for which the correct solution fails with relatively high probability or correct solution times out with relatively high probability
- There are some tasks with deliberately invalid code
- If the solutions to the tasks(Either in ReferenceImplementation or Workbook) are work_in_progress 
This PR aims to identify those tasks and exclude them.  See https://github.com/microsoft/QuantumKatas/issues/485 for more details.

@tcNickolas, I made a small change to the implementation as discussed in https://github.com/microsoft/QuantumKatas/issues/485. Instead of excluding tasks which need to be excluded from validation using <code>exclude_from_validation</code> tag; I have specified a <code>$exclude_from_validation </code> set to specify the set of cell tags which would be excluded from validation so that its easy for the contributors to know why the specific task is being excluded from validation <code>$exclude_from_validation = "['two_player_game', 'failure_of_correct_solution', 'timeout', 'invalid_code', 'work_in_progress']" </code>. And then I exclude all cell tags in <code>$exclude_from_validation </code> set by  <code>$CheckNotebook --TagRemovePreprocessor.remove_cell_tags=$exclude_from_validation</code>

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -90,6 +90,16 @@ Updating the Katas to a different QDK version can be done using PowerShell scrip
 
 After running this script you should validate that the update didn't introduce any breaking changes; see the next section for how to do this.
 
+### Excluding the tasks so that CI doesn't fail
+Currently some tasks are excluded from validation by adding the appropriate cell tag to the task due to several reasons : 
+- Some tasks require implementing two code cells at once before running the test, so the first of the cells implemented is guaranteed to fail(```two_player_game```)
+-  There are some tasks for which the correct solution fails with relatively high probability(```failure_of_correct_solution```) or correct solution times out with relatively high probability(```timeout```)
+- There are some tasks with deliberately invalid code(```invalid_code```)
+- If the solutions to the tasks(Either in ReferenceImplementation or Workbook) are work in progress(```work_in_progress```).
+
+To do this, go to ```View -> Cell Toolbar -> Tags``` to see the cells that have been tagged and to add a new tag to a cell. If there is a new reason for some tasks to be excluded from validation, feel free to update the [exclude_from_validation set](https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1#L92) in ```validate-notebooks.ps1``` script. 
+
+PS : After adding the tags to the jupyter notebook, please do ```View -> Cell Toolbar -> None``` and save your changes through ```Ctrl+S``` or by hitting ```Save and Checkpoint``` button. This is necessary because we don't want learner to see the cell tags associated with each cell when they open the jupyter notebook.
 
 ### Validating your changes
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -5,18 +5,15 @@ We're so glad you asked!
 **Table of Contents**
 
 * [Reporting Bugs](#reporting-bugs)
-
 * [Improving Documentation](#improving-documentation)
-
 * [Contributing Code](#contributing-code)
    * [Improving existing katas](#improving-existing-katas)
    * [Contributing new katas](#contributing-new-katas)
    * [Style guide](#style-guide)
    * [Updating the Katas to the new QDK version](#updating-the-Katas-to-the-new-QDK-version)
    * [Validating your changes](#validating-your-changes)
-
+      * [Excluding individual tasks from validation](#excluding-individual-tasks-from-validation)
 * [Contributor License Agreement](#contributor-license-agreement)
-
 * [Code of Conduct](#code-of-conduct)
 
 ## Reporting Bugs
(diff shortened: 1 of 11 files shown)


Manvi-Agrawal:
Hi, @tcNickolas. The CI is failing currently for task T25_PlayQuantumCHSH in CHSHGame even after increasing tolerance limit to 0.1. Seems that we might need to exclude this task or further increase the tolerance limit.

scripts/validate-notebooks.ps1
```diff
@@ -83,16 +83,24 @@ function Validate {
             </configuration>
         " | Out-File ./NuGet.Config -Encoding utf8
 
+		# Specifying tasks that need to be excluded because of various reasons :
```
tcNickolas, line 86: I would replace this comment with a link to the corresponding section of CONTRIBUTING.md - this way we won't have to maintain the explanations of the tags in two places. We'll still have to keep the *list* of tags in both places (in the script for actual validation and in CONTRIBUTING.md for explanations), but that's a lot easier.

scripts/validate-notebooks.ps1
```diff
@@ -83,16 +83,24 @@ function Validate {
...
+		# two_player_game : These tasks require implementing two code cells at once before running the test, so the first of the cells implemented is guaranteed to fail.
+		# failure_of_correct_solution : Tasks for which the correct solution fails with relatively high probability.
+		# timeout : Tasks for which the correct solution times out with relatively high probability.
+		# invalid_code : Tasks with deliberately invalid code
+		# work_in_progress : If the solutions to the tasks(Either in ReferenceImplementation or Workbook) are work_in_progress
+		$exclude_from_validation = "['two_player_game', 'failure_of_correct_solution', 'timeout', 'invalid_code', 'work_in_progress']"
```
tcNickolas, line 92: Please use spaces instead of tabs everywhere in this file

tutorials/ExploringGroversAlgorithm/ExploringGroversAlgorithmTutorial.ipynb
```diff
@@ -363,7 +371,11 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "timeout"
```
tcNickolas, line 376: I would mark the rest of the cells in this notebook as "invalid_code" as well - technically it's valid but it's clearly not the complete code for this task, so we shouldn't be validating it as such.

.github/CONTRIBUTING.md
```diff
@@ -90,6 +90,16 @@ Updating the Katas to a different QDK version can be done using PowerShell scrip
 
 After running this script you should validate that the update didn't introduce any breaking changes; see the next section for how to do this.
 
+### Excluding the tasks so that CI doesn't fail
+Currently some tasks are excluded from validation by adding the appropriate cell tag to the task due to several reasons : 
+- Some tasks require implementing two code cells at once before running the test, so the first of the cells implemented is guaranteed to fail(```two_player_game```)
```
tcNickolas, line 95: I would suggest the following list of tags, with (hopefully) more accurate tags:
* `timeout` and `invalid_code` are good
* `multicell_solution` instead of `two_player_game` (I can envision three-player-games or tasks that are not games, and technically the issue with them is that there are multiple cells that need to be executed at once). The later tasks in CHSHGame and MagicSquareGame would also be a special case of this (because they rely on earlier tasks in the notebook which we exclude from running)
* `randomized_solution` instead of `failure_of_correct_solution` (correct solution shouldn't fail, by definition,  but correct randomized solution is expected to fail and it's less scary)
* I think we don't have any `work_in_progress` right now? If we had, this seems like it would be `invalid_code`, so we could  get away with just 4 tags

tcNickolas (changes requested):
Looks great! I made a pass and left some comments. 

Also, I seem to recall that one of the tasks in PhaseEstimation kata is failing occasionally (randomized correct solution as well), I think it was 1.4 but I can't find a recent CI failure to confirm. Could you do some experimentation locally to check whether this is the case, and if yes, add it to the excludes?

Thank you!

MagicSquareGame/MagicSquareGame.ipynb
```diff
@@ -556,6 +584,7 @@
   }
  ],
  "metadata": {
+  "celltoolbar": "Tags",
```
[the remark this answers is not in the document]
Manvi-Agrawal: Yes, this setting means that learner would be able to see the tags associated with each cell which isn't required. Removed it.

CHSHGame/Tests.qs
```diff
@@ -159,6 +159,6 @@ namespace Quantum.Kata.CHSHGame {
                 set wins = wins + 1;
             }
         }
-        EqualityWithinToleranceFact(IntAsDouble(wins) / 10000., 0.85, 0.01);
+        EqualityWithinToleranceFact(IntAsDouble(wins) / 10000., 0.85, 0.9);
```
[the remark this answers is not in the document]
Manvi-Agrawal: Restored tolerance to 0.01. I can try investigating reason fo failure of Task 2.5 after resolving all the comments on this PR.

tutorials/ExploringGroversAlgorithm/ExploringGroversAlgorithmTutorial.ipynb
Manvi-Agrawal: Changed the reason from timeout to invalid_code for Exercise 4 and Exercise 5 in ExploringGroversAlgorithmTutorial.

scripts/validate-notebooks.ps1
Manvi-Agrawal: Done.

Manvi-Agrawal:
@tcNickolas . Interestingly, this time build failed due to `T06_OddChessPattern` in `UnitaryPatterns workbook`. Seems there are too many randomized solutions here :D. Is there any way to set a seed in Q# in order to minimize randomness? Like `numpy` library in Python allows us to set seed using `numpy.random.seed(0)`

.github/CONTRIBUTING.md
Manvi-Agrawal: Replaced `two_player_game` with `multicell_solution` and `failure_of_correct_solution` with `randomized_solution`. Also, removed the `work_in_progress` tag.

scripts/validate-notebooks.ps1
Manvi-Agrawal: Done

Manvi-Agrawal:
> Also, I seem to recall that one of the tasks in PhaseEstimation kata is failing occasionally (randomized correct solution as well), I think it was 1.4 but I can't find a recent CI failure to confirm. Could you do some experimentation locally to check whether this is the case, and if yes, add it to the excludes?

Hi @tcNickolas, I tried running PhaseEstimation kata a couple of times on my local, and all the tests are passing currently.

tcNickolas:
UnitaryPatterns doesn't have randomized solutions - sometimes CI build just fails for no apparent reason, I consider it an issue only it repeats several times. There is a way to set seed when defining a simulator ([the docs](https://docs.microsoft.com/en-us/quantum/user-guide/machines/full-state-simulator#seeding-the-simulator)) from a classical driver, but I'm not sure how to do this with `@Test` attribute. Generally I think it's better to keep those tests seed-less, since if the learner writes a solution that fails on the first run, with a varying seed they can retry and get a different result, but with a fixed seed they'll be stuck.

For the PhaseEstimation, transient failures are hard to repro :-) Once this change is in, I'll just keep an eye on things that fail and dig into the reasons and the frequencies, so we'll be able to keep a better track of this.

tcNickolas (approved):
Looks great! I did a bit of cleanup, rearranging the new section in the contributors guide and such.

Thank you! This change should greatly improve our CI builds coverage - and more elegantly than I imagined originally!

PR #540, opened 2020-10-20
Bring back test which are not covert in Test.qs files

vivanwin:
With the introduction of the new test methods, some test where not recognized as test anymore.


tcNickolas (approved):
Looks good. I added task numbers to several of the tasks for consistency, so that they show in the proper places in the list of tests ordered by task number.

Thank you!

PR #541, opened 2020-10-20
Trivial change.

anpaz:


tcNickolas:
The goal of this PR was to have the latest commit in the main branch that would not have `"` in the commit message. Since our CI is pretty backed up now, I'll close this PR to reduce the load, and will merge #540 instead to reach the same objective.

Issue #531, opened 2020-10-12
Add new tasks to the Truth Table kata


Manvi-Agrawal:
Add the following tasks in Truth Table kata 
- Specify truth table for 3-bit input function if exactly 1 bit out of x1, x2 or x3 is true
- Specify truth table for 3-bit input function if exactly 2 bits out of x1, x2 or x3 are true

Manvi-Agrawal:
As per discussion in https://github.com/microsoft/QuantumKatas/issues/522#issuecomment-706914502

PR #545, opened 2020-10-25
Adding new tasks to the truth table kata

Manvi-Agrawal:
Adding the following new tasks to the truth table kata as per https://github.com/microsoft/QuantumKatas/issues/531
- Specify truth table for 3-bit input function if exactly 1 bit out of x1, x2 or x3 is true
- Specify truth table for 3-bit input function if exactly 2 bits out of x1, x2 or x3 are true


tcNickolas (approved):
Looks good. I went ahead and did some cleanup to make sure the terminology we're using in the new tasks matches the one from the old tasks (for example, I think the functions we want to define in tasks 2 and 3 are not projective, that's a term used for a function that equals one of the inputs; and we don't need to specify the little endian/big endian convention in these tasks, since the order of bits doesn't matter in this case). I also added a leading 0 to the test indices in Tests.qs - since we hit 10 tasks, without that leading 0 test T10 would show right after test T1, which doesn't match the order in which we want these tests shown.

Thank you!

PR #547, opened 2020-10-26
Update README.md

TechGlue:


tcNickolas (approved):
Good point, a link to the visual explanation should help the learner.
I edited the pull request a bit to make the new link part of the list, rather than a standalone link, and to explain to the learner what it offers. 
Thank you!

PR #548, opened 2020-10-26
Update README.md

TechGlue:


tcNickolas (changes requested):
Good point, this MS Learn module covers exactly the topic pf this kata, we should have it in the list of resources.

Could you please edit the new link so that it is a part of the list of other resources (add a `*` before it and remove the empty line between it and the next list item)? It would also be nice to offer a sentence to explain what this module offers (again, similar to other links), something like "This Microsoft Learn module walks you through solving graph coloring problems using Grover's search".

You can do these edits from GitHub web interface at https://github.com/TechGlue/QuantumKatas/edit/patch-2/GraphColoring/README.md - this will add new commits to the same branch, and they will update the pull request automatically.

Thank you!

tcNickolas (approved):
Looks good, thank you!

Issue #489, opened 2020-09-27
[RandomNumberGeneration] New task: generate a random number between N and M


tcNickolas:
We can reuse task 3 and add repeat-until-success loop to discard numbers less than N or greater than M, similar to how the [Microsoft Learn module](https://docs.microsoft.com/en-us/learn/modules/qsharp-create-first-quantum-development-kit/5-random-number-generator) does it.

jainvasu631:
I would love to work on this.

tcNickolas:
I think this task might be too easy for you :-) I'm filing a bunch of easy to start with issues with an eye towards [Hacktoberfest](https://hacktoberfest.digitalocean.com/), so that if a new contributor walks in looking for something to try their hand at, they'd have something non-daunting to do.

I appreciate your enthusiasm, but I'd suggest to leave those easy issues (the new ones marked as "good first issue") as an opportunity for first-time contributors. Meanwhile I'm trying to catch up on reviewing your existing PRs so that you'd have something to do :-)

Yash-g17:
Hey can I work on this ?

tcNickolas:
@Yash-g17 Sounds good! Let us know if you need any pointers regarding the process. 

(I would recommend to send the actual pull request in October, so that it counts for Hacktoberfest :-) I've been part of this event for several years, and I like it a lot)

Yash-g17:
Tysm :)

Yash-g17:
Hi,
I have a quick question. Do we need to add a completely new package with the test suite or just add in an existing package. I can find a code file doing random number generation ReferenceImplementation.qs
Can you please guide me through this

vivanwin:
Hi @Yash-g17,

To add a task to an existing Kata you do not have to create new packages or files.
You have to add the task to the Jupyter notebook, this is the part where the user will interact with the new task. Next the actual solution should the added to the ReferenceImplementation.qs file. 
Lastly, you have to add a unit test to Test.qs, which will compare the solution given by the user in the Jupyter Notebook and the reference implementation. 

If you have anymore question don't hesitate to ask.

tcNickolas:
A quick note: there is also the Tasks.qs entry which is a different front-end for the task, that matches the operation signature in the notebook (without it the Q# project won't compile). You can find the examples in the tutorial - just look for an entry for the task 1 in each file, and see how operations are named (names are important to make sure all pieces work together) and what they do.

[...]

PR #549, opened 2020-10-27
Added task random number generation in a given range

pandasamanvaya:
Added a new task to generate a random number in a given range as mentioned in #489 .


tcNickolas (approved):
Looks good, thank you! 

I would normally suggest some experimentation with Q# to see how we can unify `CheckFlatDistributionRange` and `CheckFlatDistribution`. The logic of these two operations is exactly the same, just the ranges generated are different, and I believe that with liberal use of [Delay operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.canon.delay) and some refactoring we could use the same code path for these two. But then it should also be possible to calculate `lowRange` and `highRange` based on `min`, `max` and `nRuns`, so as to reduce the number of parameters passed to this operation. And it should be possible to re-run a test if it failed for the first time, so as to make the tests more reliable and to enable including this tutorial in the CI build... The list grows pretty long pretty fast :-) Let me file a separate issue to refactor these tests, and merge this PR as is (with just a bit of formatting cleanup, marking the test as test and other small things like that).

Thank you!

PR #556, opened 2020-11-05
Hint fix in the measurements Kata

viktorveis:
Moved the hint about an extra qubit to the right Kata: "Task 2.3**. Peres/Wooters game". "Task 2.2**.  |0⟩ ,  |+⟩  or inconclusive" doesn't need an extra qubit. Fixed a typo.

### DIFF
--- a/Measurements/Measurements.ipynb
+++ b/Measurements/Measurements.ipynb
@@ -674,13 +674,7 @@
     "\n",
     "The state of the qubit at the end of the operation does not matter.\n",
     "\n",
-    "> This task is an example of unambiguous state discrimination.\n",
-    "\n",
-    "<br/>\n",
-    "<details>\n",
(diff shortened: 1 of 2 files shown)


tcNickolas (changes requested):
Good point, I don't think we ever had a reference solution for 2.2 that relied on extra qubits, so I wonder why that hint was there... 

Would you mind doing a matching change to the [Tasks.qs file](https://github.com/microsoft/QuantumKatas/blob/main/Measurements/Tasks.qs#L317), moving the hint there as well? We try to keep the two versions of the frontend in sync for the sake of learners who choose to use the Q# project frontend.

Thank you!

viktorveis:
Sure, I updated Tasks.qs.

tcNickolas (approved):
Looks great, thank you!

PR #559, opened 2020-11-09
Update for compiler API change that removes NonNullable

samarsha:
See microsoft/qsharp-compiler#729.


tcNickolas (approved):
Looks good. Let me know if you want to merge or wait for Bettina's review?

samarsha:
I think we should probably wait to merge this until after microsoft/qsharp-compiler#729 is merged, otherwise the end-to-end build might break.

Issue #544, opened 2020-10-24
Switch from CounterSimulator.GetMaxQubitCount to AllowAtMostNQubits


tcNickolas:
**Context**

Currently we implement the tests for the tasks that restrict the number of qubits the solution can allocate using a custom simulator `CounterSimulator`, which offers [several counting operations](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Common/Utils.qs), in particular `ResetQubitCount` and `GetMaxQubitCount`. 

**Improvement**

We should be able to use the recently introduced library operation [AllowAtMostNQubits](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.allowatmostnqubits) instead, and use the regular `QuantumSimulator` instead. The affected katas are:

* [ ] GraphColoring, task 1.5
* [ ] RippleCarryAdder, task 3.5
* [ ] PhaseEstimation, tasks 2.1 and 2.2
(In task 2.1 the test also uses `GetOracleCallsCount` operation, which should be possible to replace with [AllowAtMostNCallsCA](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.allowatmostncallsca) - needs some experimentation.)

This needs some careful testing to make sure the correct solutions passe and allocating extra qubits fails.

**Example**
In GraphColoring task, for example, [task 1.5 check](https://github.com/microsoft/QuantumKatas/blob/main/GraphColoring/Tests.qs#L141) should look something like this:

```qsharp
within {
    AllowAtMostNQubits(2*N+1, "You are not allowed to allocate extra qubits.");
} apply {
    CheckColorEqualityOracle(N, ColorEqualityOracle_Nbit);
}
```

Feel free to target one test at a time!

JimmyHamel:
I Would like to work on these! 

Also, while testing GraphColoring, task 1.5 using the reference implementation i got errors. It looks like either the solution allocated twice the qubits, or the current simulator counts double the qubits. 

Not sure if im doing something wrong but im copying : 
 within {
            for ((q0, q1) in Zip(c0, c1)) {
                // compute XOR of q0 and q1 in place (storing it in q1)
                CNOT(q0, q1);
            }
        } apply {
            // if all XORs are 0, the bit strings are equal
            (ControlledOnInt(0, X))(c1, target);
        }

Inisde   operation ColorEqualityOracle_Nbit located in Task.qs

[...]

PR #560, opened 2020-11-09
Switch from CounterSimulator.GetMaxQubitCount to AllowAtMostNQubits (#544)

JimmyHamel:
Using AllowAtMostNQubits we are able to replace the usage of the QuantumCounter in some tasks to use the basic QuantumSimulator. 
Related issue: https://github.com/microsoft/QuantumKatas/issues/544

### Major changes 
**Affected Tasks**
 GraphColoring, task 1.5
 RippleCarryAdder, task 3.5
 PhaseEstimation, tasks 2.1 and 2.2

Note: at first we wanted also replace AllowAtMostNCallsCA instead of GetOracleCallsCount in task 2.1, but I found a bug in AllowAtMostNCallsCA where a controlled unitary is considered the same as its non-controlled variant. 
ref: https://github.com/microsoft/QuantumLibraries/issues/367

When this bug will get fix, we will be able to replace GetOracleCallsCount  by AllowAtMostNCallsCA  in bulk wherever it is used. 

### Minor changes
GraphColoring/Tasks.qs: added import from open Microsoft.Quantum.Arrays;. Current reference implementation needs it.
PhaseEstimation/Tasks.qs: added import from   Microsoft.Quantum.Measurement. Current reference implementation needs it.

### Testing

1. Apply change in test.qs
2. Copy reference implementation in task.qs
3. dotnet test: Verify that the test pass
4. Modify implementation in task.qs to allocate one more qubit
5. dotnet test: Verify that the test fails with correct error output.

### DIFF
--- a/GraphColoring/Tasks.qs
+++ b/GraphColoring/Tasks.qs
@@ -6,6 +6,7 @@ namespace Quantum.Kata.GraphColoring {
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Arrays;
 
     //////////////////////////////////////////////////////////////////
     // Welcome!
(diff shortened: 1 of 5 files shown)


tcNickolas (approved):
Looks great, thank you!

PR #565, opened 2020-11-17
Avoid pyzmq incompatibility in CI build

rmshaffer:
A recent release of the `pyzmq` package (20.0.0) has resulted in some type of intermittent deadlock issues when running Jupyter with IQ#. We can see this clearly in this repo, because many of the recent Katas CI builds are failing.

We haven't yet identified the root cause of this incompatibility, but we want to avoid getting that behavior in our CI builds, so this change pins the `pyzmq` version to the previous release (19.0.2).

### DIFF
--- a/scripts/steps-init.yml
+++ b/scripts/steps-init.yml
@@ -15,7 +15,7 @@ steps:
     architecture: 'x64'
   displayName: 'Use Python 3.6'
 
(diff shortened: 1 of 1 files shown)


vxfield (approved):
Tested and, although I couldn't reproduce the issue with the latest version of pyzmq, the pinned version is working fine.

tcNickolas (approved):

PR #567, opened 2020-11-19
Update Superposition.ipynb

vykrum:
Task 1.9. Superposition of all basis vectors.
Example doesn't correspond with the Goal

### DIFF
--- a/Superposition/Superposition.ipynb
+++ b/Superposition/Superposition.ipynb
@@ -330,7 +330,7 @@
     "\n",
     "**Goal:**  Change the state of the qubits to an equal superposition of all basis vectors $\\frac{1}{\\sqrt{2^N}} \\big (|0 \\dots 0\\rangle + \\dots + |1 \\dots 1\\rangle\\big)$.\n",
     "\n",
-    "> For example, for $N = 2$ the final state should be  $\\frac{1}{\\sqrt{2}} \\big (|00\\rangle + |01\\rangle + |10\\rangle + |11\\rangle\\big)$.\n",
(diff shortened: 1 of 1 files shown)


Superposition/Superposition.ipynb
```diff
@@ -330,7 +330,7 @@
     "\n",
     "**Goal:**  Change the state of the qubits to an equal superposition of all basis vectors $\\frac{1}{\\sqrt{2^N}} \\big (|0 \\dots 0\\rangle + \\dots + |1 \\dots 1\\rangle\\big)$.\n",
     "\n",
-    "> For example, for $N = 2$ the final state should be  $\\frac{1}{\\sqrt{2}} \\big (|00\\rangle + |01\\rangle + |10\\rangle + |11\\rangle\\big)$.\n",
+    "> For example, for $N = 2$ the final state should be  $\\frac{1}{\\sqrt{2^2}} \\big (|00\\rangle + |01\\rangle + |10\\rangle + |11\\rangle\\big)$.\n",
```
tcNickolas, line 333:
```suggestion
    "> For example, for $N = 2$ the final state should be  $\\frac{1}{2} \\big (|00\\rangle + |01\\rangle + |10\\rangle + |11\\rangle\\big)$.\n",
```

tcNickolas (changes requested):
Good catch! I think it would be nice to simplify this a bit and say just "1/2" instead of "1/sqrt(2^2)". Could you pleas edo this change? (You can just "accept suggestion", it will be faster)

Thank you!

tcNickolas (approved):
Thank you!

tcNickolas:
Also, I'm glad to see you've figured out how to send pull requests! :-)

vykrum:
Realised it was silly to not figure it out :-)

PR #568, opened 2020-11-26
Update Workbook_Superposition_Part2.ipynb

vykrum:
Correction in workbook corresponding to correction in Superposition kata


tcNickolas (approved):
Good catch, thank you! I remembered to check that the Q# project "backend" in Tasks.qs doesn't have this issue (it doesn't have this example at all), but missed that problem statement is duplicated in the workbook...

PR #569, opened 2020-11-26
Update QDK to version 0.14.2011120240

tcNickolas:


guenp (approved):
Looks good - optionally we can remove any of the parentheses around repeated call expressions:
> Parentheses around repeated call expressions are now optional: (Foo(x))(y) may be written as Foo(x)(y).

tcNickolas:
This release introduced a performance degradation in Q# Jupyter Notebooks - it's more noticeable in CI build times increase on "Validate Notebooks" jobs, but it also seems to happen in user-facing experience. I'll close this PR and wait for the next release which will hopefully fix this.

PR #571, opened 2020-12-07
Update Workbook_ComplexArithmetic.ipynb

JohanC68:
Some modulo confusion here. Either write:
"Python note*:  `%` is the Python modulo operator which returns the remainder of a division. For example, `7%2` gives 1, because 1 is the remainder of dividing 7 by 2."
OR
"Python note*:  `%` is the Python modulo operator which returns the remainder of a division. For example, `7%3` gives 1, because 1 is the remainder of dividing 7 by 3."


tcNickolas (approved):
Nice catch, thank you!

PR #574, opened 2020-12-09
Add %config and %trace commands to notebooks

vivanwin:


tutorials/MultiQubitSystems/MultiQubitSystems.ipynb
```diff
@@ -267,7 +267,7 @@
     "If you aren't familiar with the output of this function for single qubits, you should revisit the tutorial on [the concept of a qubit](../Qubit/Qubit.ipynb#Demo:-Examining-Qubit-States-in-Q#). \n",
     "When printing the state of multi-qubit systems, this function outputs the same information for each multi-qubit basis state.\n",
     "\n",
-    "> `DumpMachine` prints system states in little-endian format: for three qubits, $|0\\rangle$ is $|000\\rangle$, $|1\\rangle$ is $|100\\rangle$, $|2\\rangle$ is $|010\\rangle$, $|3\\rangle$ is $|110\\rangle$, and so on."
+    "> `%config dump.basisStateLabelingConvention=\"Bitstring\"` changes the `DumpMachine` format from little-endian to bitstring. For example this LittleEndian display $|3\\rangle$ this would be $|110\\rangle$ in bitstring format something we are more used to. "
```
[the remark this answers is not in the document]
tcNickolas: Could you also go through the demo and remove the little endian kets in the `Message`s? Again, they were there to help make the connection from the bitstring to the output, and if the output is the bitstring we don't need them

tutorials/MultiQubitSystems/MultiQubitSystems.ipynb
```diff
@@ -345,6 +345,7 @@
    },
    "outputs": [],
    "source": [
+    "%config dump.basisStateLabelingConvention=\"Bitstring\"\n",
```
tcNickolas, line 348: We could add a note (before or after this cell) saying that you can change the format used by the DumpMachine to bitstring, little endian or big endian using `%config` - this way the default will be bitstring but we'll teach the learner the tools to switch

tutorials/SingleQubitGates/SingleQubitGates.ipynb
```diff
@@ -404,6 +404,13 @@
     "}"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Exectuting the cell below with the `%trace` command will execute the `PauliGates_Demo` function defined above, and will print all the gates nicely at the botom of the output have a look."
```
tcNickolas, line 411:
```suggestion
    "In the previous tutorials we used `%simulate` command to run the Q# code on the full-state simulator. Here we will use `%trace` command: it will run the code on the full-state simulator as well, but will also print the circuit diagram of the run after the output."
```

tutorials/SingleQubitGates/SingleQubitGates.ipynb
```diff
@@ -342,7 +342,7 @@
    "source": [
     "### <span style=\"color:blue\">Demo: Pauli Gates</span>\n",
     "\n",
-    "The following cell contains code demonstrating how to apply gates in Q#, using the Pauli $X$ gate as an example. It sets up a series of quantum states, and then shows the result of applying the $X$ gate to each one. To run the demo, run the next cell using `Ctrl+Enter` (`⌘+Enter` on a Mac), then run the cell after it to see the output.\n",
+    "The following cell contains code demonstrating how to apply gates in Q#, using the Pauli $X$ gate as an example. It sets up a series of quantum states, and then shows the result of applying the $X$ gate to each one. To run the demo, run the next cell using `Ctrl+Enter` (`⌘+Enter` on a Mac).\n",
```
tcNickolas, line 345: I would keep this phrasing - we're adding the note to clarify the next cell but we still need to run it. This is just the 2nd tutorial in which the learner sees Q# and how it works in the notebook, so it makes sense to reiterate that you only get output after the 2nd cell.

tcNickolas (changes requested):
I made a pass, I think using bitstring output will greatly improve those initial tutorials and katas. I would also use this setting in the Superposition kata, since it uses DumpMachine extensively in error messages as well.

Thank you!

tutorials/SingleQubitGates/SingleQubitGates.ipynb
vivanwin: I have added a new cell after the code, but before the `%trace` call to reiterate how to execute the code. But if you want I can also bring this back

BasicGates/BasicGates.ipynb
```diff
@@ -30,6 +30,22 @@
     "In all tasks you need to implement the transformation exactly, without introducing any global phase. You can read more about the global phase of quantum states [here](../tutorials/Qubit/Qubit.ipynb#Relative-and-Global-Phase)."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "> Executing the below cell will change the way qubits are displayed in the tasks results, this is consistent with the notation which we use throughout the Katas."
```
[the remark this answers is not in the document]
vivanwin: This a way better solution. I have added a `.iqsharp-config.json` file with the config to the root folders, this will load the settings without having to execute any cells. As a bonus this will directly also work for the Workbooks.

tcNickolas (approved):
Looks great! I removed the little endian kets from the output in MultiQubitSystems demo, since we removed all little endian fragments from this demo, and keeping them would be confusing.

Thank you!

PR #576, opened 2020-12-14
Update Workbook_Measurements.ipynb

vykrum:
A few minor alterations


Measurements/Workbook_Measurements.ipynb
```diff
@@ -470,9 +470,9 @@
     "Like in the previous solution, we are looking for the index Q where the two bit strings differ. \n",
     "Let's define a function `FindFirstSuperpositionDiff()` which searches for an index Q which has 2 properties:\n",
     "\n",
-    "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`, and the same for all arrays in `bits2`.  \n",
+    "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`, and is the same for all arrays in `bits2`.  \n",
```
tcNickolas, line 473: This reads to me like those values should be the same in both arrays, which will get confusing once we get to item 2... How about "and the same holds for all arrays in bits1"?

Measurements/Workbook_Measurements.ipynb
```diff
@@ -450,7 +450,7 @@
     " - all the bit strings in the second array have the same value in this position (all `bits2[j][Q]` are the same),\n",
     " - these values are different for the first and the second arrays.\n",
     "\n",
-    "> For example, for arrays `[[false, true, false], [false, true, true]]` and `[[true, false, true], [false, false, true]]` return 0 corresponds to state $\\frac{1}{\\sqrt2}\\big(|010\\rangle + |011\\rangle\\big)$, return 1 - to state $\\frac{1}{\\sqrt2}\\big(|101\\rangle + |001\\rangle\\big)$, and you can distinguish these states perfectly by measuring the second qubit.\n",
+    "> For example, for arrays `[[false, true, false], [false, true, true]]` and `[[true, false, true], [false, false, true]]` return 0 corresponds to state $\\frac{1}{\\sqrt2}\\big(|010\\rangle + |011\\rangle\\big)$, return 1 to state $\\frac{1}{\\sqrt2}\\big(|101\\rangle + |001\\rangle\\big)$, and you can distinguish these states perfectly by measuring the second qubit.\n",
```
tcNickolas, line 453: I think we should keep the `-`, since it indicates that the word "corresponded" was omitted?

Measurements/Workbook_Measurements.ipynb
vykrum: Took me a while to make sense of this sentence and from what I understood (as a layman with no real background in programming or mathematics) the corresponding values in both bit strings need to be the same to satisfy this test.
I suppose my suggestion can be ignored, if it seems alright as it is.

Measurements/Workbook_Measurements.ipynb
vykrum: Again, as a layman :-), might prefer not to have symbols among words, if avoidable.
Tend to look for mathematical relevance/meaning, especially when it follows a number.

Measurements/Workbook_Measurements.ipynb
tcNickolas: 2nd item says that they have to be 0 in all elements of the first array and 1 in all elements of the second one or vice versa, but not the same across both arrays. How about splitting the 1st item into two, one for each array, and spelling it out completely? Something like "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`. If this is not the case, you cannot be sure that measuring the corresponding qubit will always return the same result. 2. The value of all arrays in `bits2` at the index Q is either `true` or `false`."

(I agree if it takes the task author some time to figure out the correct interpretation, it can be clarified :-))

Measurements/Workbook_Measurements.ipynb
tcNickolas: Then let's say "and return 1 corresponds to state ...", so that there is no word omission. 

Can you please do these two changes?

vykrum: Not sure if it's the correct approach, but created a fresh pull request with the suggested changes

tcNickolas: Typically you edit the branch from which you sent this pull request to apply any requested changes on top of the changes you proposed originally, so that in the end they all get merged together. Right now, for example, this PR contains a change not included in the other one (the escape character before % symbol on line 475). 

Looks like you've already deleted the original branch from which this PR was sent, so let's move over to #577 and apply matching changes there. I'll close this PR

tcNickolas:
Superseded by #577

Measurements/Workbook_Measurements.ipynb
vykrum: I'll keep that in mind and wait before I delete forked repositories in future.
Thanks for explaining

tcNickolas: Yes, it's better to keep the repo (and the branch in it) intact until the fate of the pull request is resolved completely :-)

No worries, it's all good things to learn.

PR #577, opened 2020-12-15
Update Workbook_Measurements.ipynb

vykrum:


Measurements/Workbook_Measurements.ipynb
```diff
@@ -470,8 +470,9 @@
     "Like in the previous solution, we are looking for the index Q where the two bit strings differ. \n",
     "Let's define a function `FindFirstSuperpositionDiff()` which searches for an index Q which has 2 properties:\n",
     "\n",
-    "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`, and the same for all arrays in `bits2`.  \n",
+    "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`.\n",
     "If this is not the case, you cannot be sure that measuring the corresponding qubit will always return the same result. \n",
+    " The value of all arrays in bits2 at the index Q is either `true` or `false`.\n"
```
tcNickolas, line 475: CI build reports this notebook as invalid JSON - I think it's because a comma is missing after the new line. I suggest the following rewording (that should also fix the broken JSON):

```suggestion
    "1. The value of all arrays in `bits1` at the index Q is either `true` or `false`, and the value of all arrays in `bits2` at the index Q is either `true` or `false`.\n",
    "If this is not the case, you cannot be sure that measuring the corresponding qubit will always return the same result. \n",
```

Measurements/Workbook_Measurements.ipynb
```diff
@@ -470,8 +470,9 @@
...
     "> For example, if you are given the state $\\frac{1}{\\sqrt2}\\big(|010\\rangle + |011\\rangle\\big)$, and if you measure the third qubit, $50\\%$ of the time you will get $0$ and $50%$ of the time $1$, therefore to get reliable information you want to measure one of the first two qubits.\n",
```
tcNickolas, line 476: We'll need to add this change from the original pull request:
```suggestion
    "> For example, if you are given the state $\\frac{1}{\\sqrt2}\\big(|010\\rangle + |011\\rangle\\big)$, and if you measure the third qubit, $50\\%$ of the time you will get $0$ and $50\\%$ of the time $1$, therefore to get reliable information you want to measure one of the first two qubits.\n",
```

tcNickolas (changes requested):
I left two comments as "suggestions" - you should be able to do matching changes using "commit suggestion", that will update your branch (and this pull request) with a new commit.

Thank you!

tcNickolas (approved):
Looks good, thank you!

PR #578, opened 2020-12-19
Maintenance fixes

vivanwin:


PhaseEstimation/Tests.qs
```diff
@@ -100,7 +100,7 @@ namespace Quantum.Kata.PhaseEstimation {
         EqualityFactI(nu, 1, $"You are allowed to call Controlled U exactly once, and you called it {nu} times");
     }
 
-    @Test("Microsoft.Quantum.Katas.CounterSimulator")
+    @Test("QuantumSimulator")
```
tcNickolas, line 103: I'm not sure this test should be switched to `QuantumSimulator`. It calls `Test1BitPEOnOnePair` which uses `GetOracleCallsCount` which is implemented only in `CounterSimulator`. Have you checked that the test still does the counting correctly, i.e., fails if the solution calls too many `Controlled U`s?

Measurements/Tests.qs
```diff
@@ -281,7 +281,7 @@ namespace Quantum.Kata.Measurements {
             [BoolArrayAsKetState(b1), BoolArrayAsKetState(b2)]);
     }
 
-    @Test("Microsoft.Quantum.Katas.CounterSimulator")
+    @Test("QuantumSimulator")
```
tcNickolas, line 284: Similarly here, I'm not sure all multi-qubit tasks should be switched to `QuantumSimulator`, since they call `DistinguishStates_MultiQubit` which uses `GetOracleCallsCount` if `measurementsPerRun > 0`. In particular, tasks 1.7 and 1.8 have a restriction on the number of measurements the solution is allowed to do. CI build won't catch that, since it only validates that correct solutions pass, not that incorrect solutions fail. Have you checked that doing extra measurements in the solution fails the tests?

tcNickolas (commented):
I did a pass - most changes look good (I did a bit of editing in a couple of places), I left comments about switch to `QuantumSimulator`. As a side note, it is helpful to leave the list of changes done in the PR in the PR description - it will make its way to the commit body during merge, and will simplify recalling what changes were done and why 6 months later :-)

Thank you!

PhaseEstimation/Tests.qs
vivanwin: I got the same doubt initially. But when I run it with the reference solution the test passes.
When I add more `Controlled U` I get the expected error message:
```
You are allowed to call Controlled U exactly once, and you called it 3 times
	Expected:	1
	Actual:	3
Try again!
```

But to be sure I can switch this back to `CounterSimulator`

Measurements/Tests.qs
vivanwin: Here I get the same as above, when adding two measurements to task 1.7 and 1.8 I get:
```
You are allowed to do at most one measurement, and you did 2
	Expected:	True
	Actual:	False
Try again!
```

I could still bring back the `CounterSimulator` to be sure.

PhaseEstimation/Tests.qs
tcNickolas: I dug deeper into this, and it turns out that IQ# (and notebooks) always use `CounterSimulator` (see https://github.com/microsoft/QuantumKatas/issues/580), but running the tests from VS/VS Code/command line will expose that failure (with a very cryptic error message, for which I'll file an issue in a moment). So please change the tests that have the operations from `CounterSimulator` back to use `CounterSimulator`, and double-check that the tests pass with the reference implementations for all tasks when executed outside of notebooks.

tcNickolas (approved):
I double-checked multi-qubit measurements tasks that don't impose a limit on the number of measurements, and they still failed tests in unexpected ways if executed on `QuantumSimulator`. I returned them to `CounterSimulator` until we fix #580.

Other than that, looks good, thank you!

PR #579, opened 2020-12-21
Validate notebook update

vivanwin:
When validating notebooks, IQSharp is installed locally in each kata folder. 
This change will install it globally, which will be nicer when validating notebooks locally.


tcNickolas:
As the first step of this change, I would see if we can just remove IQ# installation from validate-notebooks.ps1, without modifying anything else. If I'm reading [azure-pipelines.yml](https://github.com/microsoft/QuantumKatas/blob/main/azure-pipelines.yml) right, CI build job "Validating notebooks" runs [steps-init.yml](https://github.com/microsoft/QuantumKatas/blob/main/scripts/steps-init.yml), which in turn has a step [install-iqsharp.ps1](https://github.com/microsoft/QuantumKatas/blob/main/scripts/install-iqsharp.ps1) - so repeating IQ# installation on CI build agents in the notebooks validation script shouldn't be necessary.

vivanwin:
I think this would only work if, the install in `install-iqsharp.ps1` would be a global install.

tcNickolas:
I tried removing IQ# installation from validate-notebooks.ps1 in #581 and it passes the CI build; I'm following up to check whether that change will work with our internal builds.

I'll close this PR, as #581 is a more likely path to success. Thank you for trying this out!

PR #581, opened 2020-12-23
Remove IQ# install from notebook validation script

tcNickolas:
This script is used in two scenarios: on a CI build agent, the build jobs run install-iqsharp.ps1 first, so repeating IQ# install is not necessary, and when running locally, the user is likely to have IQ# installed by other means, so reinstalling it causes version conflicts down the road.


tcNickolas:
@anpaz Could you please take a look at this PR? Your main concern was that the internal E2E build relies on this script installing IQ#, so I ran it for this change [here](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=25914&view=results) to make sure it passes. And it passes the Katas CI and does not install IQ# locally, so it does what I want.

scripts/validate-notebooks.ps1
```diff
@@ -34,7 +34,14 @@ Param(
 )
 
 
-& "$PSScriptRoot/install-iqsharp.ps1"
+# Detect if we're running on Azure Pipelines.
+$IsAzurePipelines = "$Env:TF_BUILD" -ne "";
```
anpaz, line 38: We normally use the AGENT_ID variable for this.

anpaz (approved):
LGTM. The only minor change is to use the AGENT_ID variable (instead of TF_BUILD) since that's what we use in other places and I'd like to keep it consistent.

scripts/validate-notebooks.ps1
tcNickolas: Sure, I can use that one instead

PR #584, opened 2021-01-03
Fix url's to Q# documentation

vivanwin:
Q# documentation has changed the location of some documents, this will point to the updated locations.


tcNickolas (approved):
Thank you for taking care of that! I clarified a couple of the links (such as pointing to controlled functor specifically or pointing to an iterations article instead of conditional branching if that was required in the task), should be good to go.

(CI build will time out again on each notebook, but since this change doesn't impact any code, it's sufficient that notebooks validation jobs check that the notebooks are still valid JSON; if they do that, I'll merge even though CI fails.)

PR #586, opened 2021-01-05
Update build trigger

anjbur:
Add a trigger to build on `main` to keep component governance up to date.


tcNickolas (approved):
I was going to suggest to make the `always` condition `false`, so that it would skip the build if there are no code changes in the past week, but then I checked the [commit activity](https://github.com/microsoft/QuantumKatas/graphs/commit-activity) and there were exactly 2 weeks without commits over the past year, so probably not worth it :-)

Thank you!

PR #587, opened 2021-01-13
Rename "borrow" variables to avoid breaking keyword change

samarsha:
[QEP 1: Implicitly-Scoped Qubit Allocation](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md) added two new keywords, `use` and `borrow`. This is a breaking change, unfortunately, and the RippleCarryAdder kata is using `borrow` as a variable name. This PR renames it to `borrowed` to avoid the breakage.

Cc: @bettinaheim

### DIFF
--- a/RippleCarryAdder/ReferenceImplementation.qs
+++ b/RippleCarryAdder/ReferenceImplementation.qs
@@ -276,15 +276,15 @@ namespace Quantum.Kata.RippleCarryAdder {
     //////////////////////////////////////////////////////////////////
 
     // Task 4.1. N-bit subtractor
-    operation Subtractor_Reference (a : Qubit[], b : Qubit[], borrow : Qubit) : Unit is Adj {
+    operation Subtractor_Reference (a : Qubit[], b : Qubit[], borrowed : Qubit) : Unit is Adj {
         // transform b into 2ᴺ - 1 - b
         ApplyToEachA(X, b);
 
         // compute (2ᴺ - 1 - b) + a = 2ᴺ - 1 - (b - a) using existing adder
         // if this produced a carry, then (2ᴺ - 1 - (b - a)) > 2ᴺ - 1, so (b - a) < 0, and we need a borrow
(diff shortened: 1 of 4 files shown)


tcNickolas (changes requested):
I agree, that's an annoying thing to rename... Could you use "borrowBit" instead? It's kind of not borrowed yet, it's where what we will need to borrow, so "borrowBit" would be a better name

samarsha:
Sure, sounds good.

tcNickolas (approved):
Looks good, thank you!

samarsha:
@tcNickolas It looks like the validations are timing out. Do we need to update the QDK version to get the fix from microsoft/iqsharp#388?

tcNickolas:
This is really weird because we didn't update the public version of the Katas to the version that was timing out... They're still on 0.13.20102604 so there should be no difference between what's happening now and what was happening in December (and the validations passed fine back then or even just a couple days ago)

tcNickolas:
Actually, the public CI ran successfully on Jan 4 and timed out on Jan 9 - and there were literally no build-related changes on the Katas between those dates...

samarsha:
Maybe the CI build agents have changed so that they are running more slowly with the validations?

tcNickolas:
Yes, I can't think of anything else that could've changed... Let me try and see whether I can track anything down

cgranade:
> Maybe the CI build agents have changed so that they are running more slowly with the validations?

That seems consistent with my investigations on IQ# as well... I agree that there's likely an issue with the build agents.

tcNickolas:
While we are investigating the cause of the timeout (unrelated to this change), I'll make an exception and merge this change without full CI passing to unblock further work on https://github.com/microsoft/qsharp-compiler/pull/815. 

This change modifies 4 files; 3 .qs files are covered by "Validate C# Projects" jobs, and I verified the change in .ipynb file manually, so it is unlikely to introduce any new issues.

PR #590, opened 2021-01-29
Update to 0.15 syntax, batch 1

tcNickolas:
This change updates the katas and tutorials to the new syntax introduced in Q# 0.15:
* "use" keyword instead of "using"
* no round brackets in for loops


vivanwin (approved):
This look good, thanks for the work.

Issue #552, opened 2020-10-28
Measurement counting mechanism should be updated for target package feature.


swernli:
When the feature work related to https://github.com/microsoft/qsharp-runtime/issues/249 is merged into main, the behavior of `M` will change such that each call to `M` ends up calling `Measure` underneath. This means that places like [Tests.qs](https://github.com/Microsoft/QuantumKatas/blob/main/Measurements/Tests.qs#L190) that validate measurement counts by adding together the number of calls to `M` and the number of calls to `Measure` will incorrectly double count and reject valid solutions. The fix is to update the counting logic to only count calls to `Measure`, but that change should only be made once the katas are updated to target a version of the QDK that includes the target package feature.

swernli:
Per conversation with @tcNickolas we think the best way to address this is fix the counting mechanism right when the target package feature gets merged. That will have correct behavior for anyone using beta QDK while being slightly over permissive for anyone using the release QDK until the feature makes it into a release. No one should experience any failures during that time, but some invalid kata solutions (that use too many calls to `M`) will be accepted, which is unfortunate but not blocking.

@bettinaheim  for FYI

swernli:
The work for https://github.com/microsoft/qsharp-runtime/issues/249 has completed the feature branch has been merged, so it's time to circle back to this issue and fix the katas as described above. I'll open a PR soon.

PR #592, opened 2021-01-29
Update Measurement counting logic

swernli:
With the feature work from https://github.com/microsoft/qsharp-runtime/issues/249 merged, the beta packages will now report oracle counts differently. This is expected, as `M` calls into `Measure`, so counting both operations will double count measurements and fail validation of solutions. This adjusts to only count `Measure` in order to ensure the right behavior in the pipeline and when the feature makes it into a public release.

This fixes Measurement counting mechanism should be updated for target package feature. #552


tcNickolas (approved):
Looks good; we'll be a little more permissive with these tasks until the release that includes this feature, but I think that's fine.

Thank you!

PR #594, opened 2021-01-30
Finish update to 0.15 syntax

tcNickolas:
This change also re-balances notebooks validation jobs to split notebooks more evenly between the two jobs.


vivanwin (approved):
Looks good, thank you.

PR #595, opened 2021-02-02
[Quickref] Update to 0.15 syntax

vivanwin:


quickref/qsharp-quick-reference.tex
```diff
@@ -586,7 +586,7 @@
 
   \begin{keysref}{Control Flow}
     Iterate over \newline a range of numbers
-                       & \texttt{for (\emph{index} in \emph{range}) \{ }\newline 
+                       & \texttt{for \emph{index} in \emph{range} \{ }\newline 
                          \texttt{\hphantom{....}// Use integer \emph{index}} \newline
                          \texttt{\hphantom{....}...} \newline
                          \texttt{\}} \newline
```
tcNickolas, line 592: Please update the syntax in `e.g.` sections as well, since both for loops have syntax examples for them

quickref/qsharp-quick-reference.tex
```diff
@@ -663,17 +663,15 @@
...
-                           \texttt{\hphantom{....}//} Qubits in \texttt{\emph{reg}} start in $\ket{0}$. \newline
-                           \texttt{\hphantom{....}...} \newline
-                           \texttt{\hphantom{....}//} Qubits must be returned to $\ket{0}$. \newline
-                           \texttt{\}} \\
-    Allocate one qubit   & \texttt{using (\emph{one} = Qubit()) \{ ... \} } \\
+                         & \texttt{use \emph{reg} = Qubit[\emph{N}]; } \newline
+                           \texttt{//} Qubits in \texttt{\emph{reg}} start in $\ket{0}$. \newline
+                           \texttt{...} \newline
+                           \texttt{//} Qubits must be returned to $\ket{0}$. \\
+    Allocate one qubit   & \texttt{use \emph{one} = Qubit();  ...  } \\
```
tcNickolas, line 670: Can you please move `...` to a new line? It looks a bit odd on the same line with `use`

tcNickolas (changes requested):
Looks good, just a couple of small fixes.

I wonder if we should also remove round brackets from places where they are optional but not deprecated? We kept them when updating the Katas code to the new syntax because it's a bit painful to track them down without a warning, but the quick ref should probably have the syntax for new code written, not old code maintained...

Thank you!

vivanwin:
I think you are right. I have removed the optional parentheses in the `if` conditions and removed the `for` parentheses in the `e.g` part.

tcNickolas (approved):
Thank you!

PR #596, opened 2021-02-03
Added ability to disable exception printing in CounterSimulator

DmitryVasilevsky:
Do not merge! Example only.
This is an example of how to make error messages in Test Explorer UI cleaner.
CounterSimulator exposes ability to disable exception printing to Q#.
Example of how to use this ability via apply-within block to print clean message in DistinguishUnitaries.
The example is commented out because Katas package needs to be updated first.

### DIFF
--- a/DistinguishUnitaries/Tests.qs
+++ b/DistinguishUnitaries/Tests.qs
@@ -79,6 +79,14 @@ namespace Quantum.Kata.DistinguishUnitaries {
         }
         // This check will tell the total number of failed classifications
         Fact(totalMisclassifications == 0, $"{totalMisclassifications} test runs out of {nTotal} returned incorrect state.");
+
+        // Use the following within-apply pattern to display clean message without
+        // additional information from exception (such as stack trace).
+        //
+        // within { DisableExceptionPrinting(); } apply {
+        //    Message($"{totalMisclassifications} test runs out of {nTotal} returned incorrect state.");
+        //    Fact(totalMisclassifications == 0, $"Failure.");
+        // }
     }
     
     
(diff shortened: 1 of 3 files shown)


tcNickolas:
This is a neat trick! 

I'm reluctant to adopt it, though, as it would require using CounterSimulator in all the katas for uniformity of their output (there are some katas that only use CounterSimulator for one test, because all other needs are addressed by the standard simulator). We are trying to do the opposite, to get away from using CounterSimulator by using built-in Q# features (and, when needed, by extending Q# features with new ones like [AllowAtMostNQubits](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.allowatmostnqubits)). This way we can use the same testing and tooling ideas for other simulators, not just the full-state one, and in addition our users can have similar tools for their projects without necessarily relying on the Katas-specific infrastructure. 

The original issue filed in https://github.com/microsoft/qsharp-runtime/issues/156 has been fixed by [adding DisableExceptionPrinting](https://github.com/microsoft/QuantumKatas/blob/4691c7689e6224e3cd10aed15c5a65f8824da67b/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L173) to the `%kata` magic, so now the exception stack trace only shows up in VS/VS Code test output. It seems a lot less disruptive there compared to the original issue in the notebooks, I don't recall hearing any user feedback about that (as opposed to the original issue!).

Let me close this PR and keep this capability in mind in case in the future we decide to switch to CounterSimulator in all katas or to return the C# driver for the test code (that's a different place where we could call DisableExceptionPrinting with similar results).

Thank you for putting together this prototype!

PR #603, opened 2021-03-05
Update NuGet.config

anjbur:
Add `<clear />` to NuGet.config to satisfy compliance requirements.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -11,8 +11,7 @@
 -->
 <configuration>
   <packageSources>
-    <clear />
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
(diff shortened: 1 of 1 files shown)


tcNickolas:
The comment on the PR doesn't match the actual change (`<clear />` was present in the first iteration but not in the second one). I'm not sure what the compliance requirements are, could you please clarify?

anjbur:
@tcNickolas Thank you for pointing that out, fixed. There's a requirement that NuGet.config files have `<clear />` present, so that if there are multiple feeds used they have to be explicitly listed.

tcNickolas (approved):
Looks good, thank you!

PR #605, opened 2021-03-11
Clean up built projects to save space in pipeline

swernli:
This matches the change in https://github.com/microsoft/Quantum/pull/464 and will help avoid failures where the build pipeline fails because it runs out of space.


tcNickolas (changes requested):
Minor comment, but could you please match the indentation in the new code to the indentation in the rest of the file? Thank you!

tcNickolas (approved):

PR #606, opened 2021-03-12
Create qdk-sync.yml

anjbur:
Create GitHub action to sync updates back to the main `qdk` repo


tcNickolas (approved):
Looks good to me; I think we should merge this PR after the secret is updated, so that we can observe the result right away without waiting for another change on the Katas.

Thank you!

tcNickolas:
Looks good - let me merge this now, and I'll try to keep in mind that we need to double-check that the action was triggered the next time we update that folder.

Thank you!

PR #609, opened 2021-03-24
Fix broken documentation links

anjbur:


quickref/qsharp-quick-reference.tex
```diff
@@ -644,8 +644,8 @@
   \section{Resources}
   
   \begin{keysref}{Documentation}
-    Quantum \newline Development Kit & \url{https://docs.microsoft.com/quantum} \\
-    \qs~Language \newline Reference  & \url{https://docs.microsoft.com/quantum/language} \\
+    Quantum \newline Development Kit & \url{https://docs.microsoft.com/azure/quantum} \\
```
tcNickolas, line 647: Please don't modify this file or re-generate the quickref/qsharp-quick-reference.pdf that is generated from it. The setup for PDF generation is a bit tricky (I don't have it on my laptop), so I'm fine with not modifying it now - I noted this need and will get back to it when we're not in a rush.

tcNickolas (changes requested):
There is also another broken link that @Bradben pointed out in  tutorials/QuantumClassification/InsideQuantumClassifiers.ipynb: after yesterday's update: https://docs.microsoft.com/azure/quantum/user-guide/libraries/machine-learning/introduction should actually be https://docs.microsoft.com/azure/quantum/user-guide/libraries/machine-learning/intro. (It might also be present in other places in that tutorial as well.)

Could you fix that one as well, since you have the PR open anyways? Thank you!

quickref/qsharp-quick-reference.tex
anjbur: Understood, change reverted!

tcNickolas (approved):
Thank you!

PR #610, opened 2021-03-30
[quick ref] Update doc links

vivanwin:


tcNickolas (approved):
Thank you for taking care of that!

PR #612, opened 2021-04-23
Fix broken link

anjbur:


tcNickolas (approved):
Thank you!

tcNickolas:
I just realized that this link is also present in the beginning of GroversAlgorithm.ipynb. Could you please do a matching change there?

tcNickolas (approved):
I went ahead and fixed the other instance of the same link. Thank you!

PR #613, opened 2021-04-25
Fix link to "Q# documentation on quantum states"

jrskerrett:
The link to "Q# documentation on quantum states" in the Qubit tutorial is incorrect. It has a slash where a dash should be.


tcNickolas (approved):
Thank you for catching this!

Issue #166, opened 2019-08-16
%kata magic doesn't recognize C# tests


tcNickolas:
PhaseEstimation kata has a test that covers Q# code but is written in C# due to the need to catch and process exceptions thrown by the task. When the kata is approached as Q# project, the test is recognized and executed together with Q# tests just fine. However, when the kata is converted to Jupyter Notebook format, the `%kata` magic command does not recognize the C# test. This is not unexpected, but ideally we need to come up with a way to execute C# tests from Q# notebooks.

This is related to one of the work items in #138.

Manvi-Agrawal:
Is the magic command %kata specified in this repository(ie Quantum Katas) or some other repository?

tcNickolas:
It's defined in [`Microsoft.Quantum.Katas` package](https://github.com/microsoft/QuantumKatas/tree/master/utilities/Microsoft.Quantum.Katas) in this repo.

Manvi-Agrawal:
Thanks @tcNickolas.

Manvi-Agrawal:
@tcNickolas I would like to take up this issue.

tcNickolas:
Sure, give it a try and let us know what design you come up with. It's a rather non-trivial issue, so we'll probably want to discuss it with @rmshaffer as well, who is familiar with IQ# internals.

Thank you!

Manvi-Agrawal:
Hi @tcNickolas @rmshaffer, do you have any idea about where to see the Logger messages defined in `KataMagic.cs`?

Some examples are 
- [`Logger?.LogWarning(e, "Unexpected error.");` ](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L149)
- [`Logger.LogDebug($"Registering {answer.Key.FullName}");`](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L178)
- [`Logger.LogDebug($"Resolved {userAnswer.FullName} to {skeletonAnswer}");`](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L223)

rmshaffer:
[...]

Issue #580, opened 2020-12-21
`%kata` magic should use the simulator specified in `@Test` attribute


tcNickolas:
Currently executing a cell with `%kata` or `%check_kata` selects the simulator to use in [this code](https://github.com/microsoft/QuantumKatas/blob/5eb28be678c3bb1cd819a212545ca51657c60481/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L206) - which always uses `CounterSimulator`. This works for now, since all tests are configured to use either `CounterSimulator` or `QuantumSimulator` from which the former is derived, but this will break if we ever define tests that use other simulators.

We should replace this code (and the [matching code](https://github.com/microsoft/QuantumKatas/blob/5eb28be678c3bb1cd819a212545ca51657c60481/utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs#L200) in `%check_kata` magic implementation) with logic that would find the `@Test` attribute, read the simulator configuration from it and use that simulator.

Manvi-Agrawal:
Hi @tcNickolas, a couple of questions to understand the scope of this issue :
- Can [custom simulators](https://devblogs.microsoft.com/qsharp/build-your-own-q-simulator-part-1-a-simple-reversible-simulator/) be used by using `@Test` attribute? Is it in scope of this PR?
- For [`ToffoliSimulator`](https://docs.microsoft.com/en-us/azure/quantum/user-guide/machines/toffoli-simulator):
Currently, it supports [StackTraceDisplay](https://github.com/microsoft/iqsharp/blob/main/src/Jupyter/Extensions.cs#L154) functionality. But I am not sure if the tests using  `ToffoliSimulator` would need to display diagnostic information using `DumpMachine` and `DumpRegister` callable
- For [`ResourcesEstimator`](https://docs.microsoft.com/en-us/azure/quantum/user-guide/machines/resources-estimator) and [`QCTraceSimulator`](https://docs.microsoft.com/en-us/azure/quantum/user-guide/machines/qc-trace-simulator/) : 
They currently support [StackTraceDisplay](https://github.com/microsoft/iqsharp/blob/main/src/Jupyter/Extensions.cs#L154) functionality. I think it might be fine for this scenario since these simulators provide analysis of the resources needed to run a quantum algorithm. It should make less sense for tests using these simulators to call `DumpMachine` and `DumpRegister` callable.

tcNickolas:
[...]

PR #614, opened 2021-04-27
Added section for contributing to Magics

Manvi-Agrawal:
This is as per the following comments : https://github.com/microsoft/QuantumKatas/issues/166#issuecomment-827118778, https://github.com/microsoft/QuantumKatas/issues/166#issuecomment-783879320 and https://github.com/microsoft/QuantumKatas/issues/166#issuecomment-802951251


Manvi-Agrawal:
Hi @tcNickolas. Changed the instructions in [Contributing Guide](https://github.com/microsoft/QuantumKatas/blob/main/.github/CONTRIBUTING.md) a bit to avoid skipping the auto-loading of projects by IQ# kernel. I think this is a slightly better technique to test the changes locally since the end goal is to ensure IQ# is successfully auto-loading the projects. Also, this technique allows to manipulate cache assembly in a bit more convenient way from the `KataMagic` and `CheckKataMagic`, should we decide to do in future.

.github/CONTRIBUTING.md
```diff
@@ -159,6 +160,47 @@ This can happen for several reasons:
 To exclude a task from validation, open the corresponding Jupyter notebook and choose ```View -> Cell Toolbar -> Tags``` to see and edit the tags for each cell. Add the tag that is the most fitting description of the failure cause to the cell. 
 After you are done with editing the notebook, choose ```View -> Cell Toolbar -> None``` to turn off tags editing view for the subsequent users of this notebook. Finally, save the notebook.
 
+#### Validate changes to `%kata` magic and `%check-kata` magic on local machine
+1. Add logging statements to the [`KataMagic`](./utilities/Microsoft.Quantum.Katas/KataMagic.cs) and [`CheckKataMagic`](./utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs).
```
tcNickolas, line 164: Since this file is not in the repo root but in `.github` folder, the relative links need to go one level up, starting with `./../` rather than just `./` (later in the change as well)

.github/CONTRIBUTING.md
```diff
@@ -159,6 +160,47 @@ This can happen for several reasons:
...
+   ```
+   Logger.LogDebug($"Value : {val}");
+   ```
+2. Generate a custom nuget package
+   1. Build the Microsoft.Quantum.Katas project to produce a NuGet package to get a version other than 1.0.0 by adding the following code in [this file](./utilities/Microsoft.Quantum.Katas/Microsoft.Quantum.Katas.csproj) by adding the following code
+   ```
+   <RootNameSpace>Microsoft.Quantum.Katas</RootNameSpace>
+   <Version>version-number</Version> ).
```
tcNickolas, line 172:
```suggestion
   <Version>version-number</Version>
```

.github/CONTRIBUTING.md
```diff
@@ -159,6 +160,47 @@ This can happen for several reasons:
...
+   <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+   ```
+   2. Copy the generated .nupkg file to the folder in which your project resides (like PhaseEstimation kata) and remove the following line from your project's project file.
+   ```
+   <PackageReference Include="Microsoft.Quantum.Katas" Version="0.16.2104138035" />
```
tcNickolas, line 177: If we mention the package version here, we'll need to update version update script to cover this line. (You'll see that [right now](https://github.com/microsoft/QuantumKatas/blob/35c153fe5b4901b0a822813ddaa77d4aeedcf165/scripts/Update-QDKVersion.ps1#L94) we update another line in this script, but not this one). I'd suggest to avoid this (especially since here the version number can be pretty much anything) and rephrase this as "Remove the package reference to Microsoft.Quantum.Katas from your .csproj file"

.github/CONTRIBUTING.md
```diff
@@ -159,6 +160,47 @@ This can happen for several reasons:
...
+   <?xml version="1.0" encoding="utf-8"?>
+   <configuration>
+     <packageSources>
+       <add key="Local Folder" value="." />
+     </packageSources>
+   </configuration>
+   ```
+3. To observe your changes in console window where Jupyter Notebook was launched.
+   1. Set the environment variable `IQSHARP_LOG_LEVEL='Debug'`.
+   > Note : The environment variable is case-sensitive.
```
tcNickolas, line 190: I think the notes and the samples here (and above as well) should be indented more to render as part of second-level list item rather than of first-level. This should make it more readable

.github/CONTRIBUTING.md
```diff
@@ -159,6 +160,47 @@ This can happen for several reasons:
...
+   ```
+   $ start jupyter notebook
+   ```
+   > This would launch Jupyter Notebooks server in a new window which would be
+   > helpful in debugging the magics since logging is a bit noisy.
+
+4. Changes to be done in jupyter notebook 
+   1. Use `%package` magic to load the newly built nuget package `Microsoft.Quantum.Katas` by adding the following code
+   ```
+   %package Microsoft.Quantum.Katas::<package-version>
```
tcNickolas, line 201: Do we need this line, or would it work if we just change the version in the `<PackageReference Include="Microsoft.Quantum.Katas" Version="0.16.2104138035" />` dependency in the .csproj file? Trying to figure out how to minimize the number of steps

tcNickolas (changes requested):
I did a quick pass, I still have to try these steps myself to make sure they work, but I left some comments independently of that.

Thank you!

.github/CONTRIBUTING.md
Manvi-Agrawal: I initially tried that approach only but the IQ# kernel for no particular reason loaded the katas package of the same version as the IQ# kernel. I think this might be due to the fact that IQ# kernel would attempt to load dependencies of the same version as the kernel if <IQSharpLoadAutomatically> property is set to true.

.github/CONTRIBUTING.md
Manvi-Agrawal: Done

.github/CONTRIBUTING.md
Manvi-Agrawal: Done

.github/CONTRIBUTING.md
Manvi-Agrawal: Done

Manvi-Agrawal:
Hi @tcNickolas. I have made the desired changes.

And as for this one :
> Do we need this line, or would it work if we just change the version in the `<PackageReference Include="Microsoft.Quantum.Katas" Version="0.16.2104138035" />` dependency in the .csproj file? Trying to figure out how to minimize the number of steps.

I initially tried that approach only but the IQ# kernel for no particular reason loaded the katas package of the same version as the IQ# kernel. I think this might be due to the fact that IQ# kernel would attempt to load dependencies of the same version as the kernel if `IQSharpLoadAutomatically` is set to true.

Manvi-Agrawal:
Hi @tcNickolas . I noticed a very strange phenomenon after qdk-update to the the version `0.16.2104138035`

if we do the following in NuGet.config as per step 2.3
```
<add key="Local Folder" value="." />
```
- In the version `0.16.2104138035`
Custom nuget now needs to be placed in the `QuantumKatas` folder(ie the same folder as that of `NuGet.config` file) 

- With early versions : `0.15.2101125897` and  `0.15.2103133969`
 We needed to move the generated nuget to the project in which the project resided.

Maybe @rmshaffer could help in confirming if changes in IQ# kernel are done or there is some problem with my local :-(

tcNickolas:
Sadly, Ryan is no longer with our team, so @anjbur is picking up IQ# ownership.
@anjbur, were there any changes that could cause this behavior change in the past month?

Generally I would suggest recommending the person working with the custom Microsoft.Quantum.Katas package to put both NuGet.config and the custom package to the folder of the project they're using as a test, rather than in folders closer to root. That's what I do myself, to limit the scope of the custom settings and to make it harder to get unintended consequences when working with other katas.

Manvi-Agrawal:
Apologies for my previous comment. Initially when I was following step 3 as mentioned [here](https://github.com/microsoft/QuantumKatas/issues/166#issuecomment-783879320), I changed the `NuGet.config` file present in the root folder of the `QuantumKatas` and added `<add key="Local Folder value="." />` and things seemed to go pretty well in `0.15.2101125897` and `0.15.2103133969` until the next Qdk-update to version `0.16.2104138035`.

As of now, these instructions work perfectly on my local machine. My only concern is that if someone would attempt to run the run the validate notebook script for the first time after creating the custom package, then the check would fail at `%package Microsoft.Quantum.Katas::<version-number>` since we explicitly clear the NuGet cache before validating the change. So, I could possibly update the instructions to change [the line here](https://github.com/microsoft/QuantumKatas/blob/main/scripts/validate-notebooks.ps1#L88) to `<add key=""Local Folder"" value=""."" />` . However if someone is likely to introduce a major change as in [Issue#580](https://github.com/microsoft/QuantumKatas/issues/580), I think they would be doing changes to the `KataMagic` first and then doing changes to the `CheckKataMagic`, so this situation won't arise.

tcNickolas (approved):
I finally got the time to sit down and go through all the debug steps - they look great, I think the only thing I didn't need was the root namespace. I did a bit of editing to call out some things that got me confused (for example, that removing package reference to M.Q.Katas will break the build but that's ok).

Thank you!

PR #620, opened 2021-05-05
[Utility magics]`%kata` magic should use simulators from `@Test` attribute

Manvi-Agrawal:
Fixes https://github.com/microsoft/QuantumKatas/issues/580

**Design document Important sections**
- [Features Planned](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit#heading=h.a3rudj7z507x)
- Changes to [`KataMagic`](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit#heading=h.pwnykvnfiwqh) and [`CheckKataMagic`](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit#heading=h.twwkdrtcquqf)
- [Refractoring Magics](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit#heading=h.bof0pq2jz1kq)
- [Testing changes locally](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit#heading=h.uihaq4qqsknb)

**Complete document:** [doc-link](https://docs.google.com/document/d/1APDB3G9mmfx7aJEe8zd7gm9XUv9JK4nCEzB7O3wbBwI/edit?usp=sharing)


utilities/Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -228,5 +233,138 @@ public virtual OperationInfo FindSkeletonAnswer(OperationInfo test, OperationInf
...
+                }
+
+            }
+            return testSimulators;
+        }
+
+        /// <summary>
+        /// Returns the list of simulator names specified via @Test attribute
+        /// </summary>
+        public List<string> GetSimNamesFromTestAttribute(OperationInfo test)
```
anjbur, line 324: It would be good to avoid the duplicated code between KataMagic and CheckKataMagic where possible. Creating a shared superclass would be a great way to reduce the duplication.

anjbur (changes requested):
Thank you @Manvi-Agrawal for implementing this! 

I ran some initial validations and the main issue is that `CreateSimulators` is always returning an empty list in CheckKataMagic. In addition to fixing that bug, if no simulation is run then there should be an error, rather than defaulting to printing a success statement.

Otherwise, the core logic to create various simulators looks good. The rest of the comments are largely focused on maintainability rather than logic. Once that `CreateSimulators` bug is fixed we can do another round of validations before looking at merging this in. Thanks again for taking a look at expanding this functionality!

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
```diff
@@ -162,21 +172,29 @@ public virtual bool Simulate(OperationInfo test, string userAnswer, IChannel cha
...
+
+                    qsim.DisableLogToConsole();
+                    qsim.Register(skeletonAnswer.RoslynType, referenceAnswer.RoslynType, typeof(ICallable));
 
-                var value = test.RunAsync(qsim, null).Result;
+                    qsim.OnLog += (msg) =>
+                    {
+                        simHasWarnings = msg?.StartsWith("[WARNING]") ?? simHasWarnings;
+                        if(simHasWarnings == true)
+                            hasWarnings = true;
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
```diff
@@ -221,5 +226,125 @@ public virtual bool Simulate(OperationInfo test, string userAnswer, IChannel cha
...
+        /// <summary>
+        /// Creates the instance of the simulator(s) on which
+        /// the test operation is to be run
+        /// These simulators are read from the configuration in @Test attribute
+        ///
+        /// [TODO] Support custom simulator(s) in Q# project
+        /// </summary>
+        public virtual List<SimulatorBase> CreateSimulators(IChannel channel, OperationInfo test)
+        {
+            List<SimulatorBase> testSimulators = new List<SimulatorBase> ();
```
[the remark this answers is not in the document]
Manvi-Agrawal: Resolved the bug

utilities/Microsoft.Quantum.Katas/Extensions.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.SyntaxTokens;
+
+namespace Microsoft.Quantum.Katas
+{
+    internal static class Extensions
+    {
+        internal static bool TryAsStringLiteral(this TypedExpression expression, out string? value)
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

utilities/Microsoft.Quantum.Katas/KataMagic.cs
Manvi-Agrawal: Great idea. I was thinking of `UtilityMagic` as a name for the superclass. Does that work or do you have another name in mind?

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
```diff
@@ -162,21 +172,28 @@ public virtual bool Simulate(OperationInfo test, string userAnswer, IChannel cha
...
-                var value = test.RunAsync(qsim, null).Result;
+                    qsim.DisableLogToConsole();
+                    qsim.Register(skeletonAnswer.RoslynType, referenceAnswer.RoslynType, typeof(ICallable));
 
-                if (qsim is IDisposable dis) { dis.Dispose(); }
+                    qsim.OnLog += (msg) =>
+                    {
+                        simHasWarnings = msg?.StartsWith("[WARNING]") ?? simHasWarnings;
+                        if(simHasWarnings == true) { hasWarnings = true; }
+                        channel.Stdout(msg);
```
Manvi-Agrawal, line 190: @tcNickolas @anjbur , it seems that code here check for any warnings in the reference implementation like Deprecated syntax. Is this needed to ensure that only the latest syntax is being used? Or could I possibly remove the code here and just do a `sim.OnLog += channel.Stdout;` .

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
```diff
@@ -162,21 +172,29 @@ public virtual bool Simulate(OperationInfo test, string userAnswer, IChannel cha
...
+                    };
 
-                if (qsim is IDisposable dis) { dis.Dispose(); }
+                    var value = test.RunAsync(qsim, null).Result;
 
+                    if (qsim is IDisposable dis) { dis.Dispose(); }
+                }
                 return !hasWarnings;
             }
             catch (AggregateException agg)
```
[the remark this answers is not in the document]
Manvi-Agrawal: @anjbur Did some changes regarding that. 
- First, ensured that `CreateSimulators` should never return an empty list. 
  I initially omitted that check since any problem in simulator creation will throw an exception. But this creates a problem if the simulatorList isn't populated after successful creation of the simulators. This might happen in the present scenario if the simulator created is not of the type `SimulatorBase`. Also, a second check wouldn't hurt much  because the impact of such a bug would be high since no tests could be validated because of this bug.
- Secondly, I also added some simulator specific messages.

utilities/Microsoft.Quantum.Katas/KataMagic.cs
anjbur: `CheckKataMagic` and `KataMagic` are related but there isn't a clear hierarchical relationship, so best to give them a common parent. Let's follow the same pattern used in iqsharp and name it `AbstractKataMagic` (see `AbstractMagic` from iqsharp [here](https://github.com/microsoft/iqsharp/blob/main/src/Jupyter/Magic/AbstractMagic.cs)).

Manvi-Agrawal: Sounds great

Manvi-Agrawal: Done the refractoring

Manvi-Agrawal:
@anjbur, I have done the changes as requested by you. 
Maybe I could remove the `SetDisplay` of the `CheckKataMagic` if the code there is not much important. 
It seems to me that we are catching warnings in the ReferenceImplementation(one of the scenarios maybe to check if any deprecated syntax is being used).

Manvi-Agrawal:
Following is the list of notebooks affected by the new change
- DeutschJozsaAlgorithm.ipynb
- DeutschJozsaAlgorithmTutorial.ipynb
- GroversAlgorithm.ipynb
- MagicSquareGame.ipynb
- Measurements.ipynb
- Oracles.ipynb
- PhaseEstimation.ipynb
- QEC_BitFlipCode.ipynb
- SingleQubitSystemMeasurements.ipynb
- SuperdenseCoding.ipynb
- TruthTables.ipynb

tcNickolas:
It is important to understand what changes in the code caused changes in
validation outcomes, before we make decisions such as excluding tasks from
validation or modifying the code. This change should only add use of new
simulators, but not break anything in the validation of the existing tasks
or remove any functionality. %kata magic always gave errors if the code was
invalid, but those were benign, they didn't fail notebook validation
because notebooks are not validated using %kata. %check_kata magic
shouldn't give errors for invalid code in the cell itself, since it doesn't
use that code.

Manvi-Agrawal:
@tcNickolas will look for the changes that caused this. Meanwhile, do we want to retain the Prototype Kata in #621?

Manvi-Agrawal:
> It is important to understand what changes in the code caused changes in
validation outcomes, before we make decisions such as excluding tasks from
validation or modifying the code. 

Hi @tcNickolas @anjbur, I am able to promote code reuse by `AbstractKataMagic` now. The core functionality of `KataMagic` and `CheckKataMagic` is retained now. I did some validation for the `KataMagic` and `CheckKataMagic` using `BasicGates` and `Prototype` notebook and then ran a custom script to check the changes in `CheckKataMagic` locally for all notebooks that have `%kata` magic.
The main point of difference was that `userAnswer` was of the type
- `OperationInfo` in `KataMagic`
- `string` in `CheckKataMagic`

I did the following changes:
- Changed the signature from `AbstractKataMagic` to `AbstractKataMagic<T>`.
- This generic type `T`, allows `userAnswer` to switch between `string` and `OperationInfo` with ease. Also, an important point here to note is that usually we were usually using the string form of `userAnswer` which I get from `userAnswer.ToString()`. Also, note that OperationInfo overrides the `ToString()` method to return `userAnswer.FullName`
- Also, the new `FindReferenceImplementation` uses the namespace of `skeletonAnswer` rather than `test`. This change doesn't have much impact because currently the `skeletonAnswer`, `referenceAnswer` and the `test` are in the same namespace.

tcNickolas:
Validating this functionality is an interesting challenge, related to retaining the prototype kata. The issue #580 went unnoticed for a very long time, until we realized that the test code that should be failing on `QuantumSimulator` was passing. So to validate that the magics pick up the right simulators we need to reproduce the situation in which we noticed that they don't. 

To figure out the set of things that need to be validated, we need to 

1. Check all simulators from the set {`QuantumSimulator`, `CounterSimulator` and `ToffoliSimulator`} and create tests that pass on each of the simulators but fail on others.
2. Run these tests on the simulators on which they should pass, and check that they pass indeed.
3. Run these tests on the simulators on which they _shouldn't_ pass, and check that they fail indeed.
4. Optionally, create tests that should run on two out of three.

The tests in the prototype kata from #621 don't actually use any functionality of the simulators, so they will pass as long as any simulator picks them up. So the test kata needs to be improved to contain non-trivial tests.

* Tests that pass on `CounterSimulator` only should use counting operations such as `GetOracleCallsCount` and `ResetOracleCallsCount` - the other two simulators don't define them, so those tests should fail at runtime.
* Tests that pass on `QuantumSimulator` but not on `ToffoliSimulator` should use gates other than X, CNOT and controlled variants of those - a Hadamard will fail immediately, since it's not supported by `ToffoliSimulator`. (I think we won't be able to do a test that passes on `QuantumSimulator` but fails on `CounterSimulator`, since the latter offers all functionality of the former.) 
* Tests that pass on `ToffoliSimulator` only should allocate and manipulate a lot of qubits, like a few hundred - full state simulator should run out of memory trying to do this, though I'm not sure how it will look in notebooks, most likely a timeout.

Does that sound reasonable? @anjbur, any other thoughts? I imagine those tests would have to be non-automated now, and hopefully we'll be able to automate them later.

tcNickolas:
Also, I don't think we need to use ResourcesEstimator and QCTraceSimulator at this point - I can imagine tests that use ToffoliSimulator easily, but these two don't have any effect noticeable from Q# code. Any tests using them would be written in C# or Python, estimating some Q# code and analyzing the results outside of Q#.

Manvi-Agrawal:
Hi @tcNickolas , I have done some changes to include non-trivial tests as suggested by you in [this comment](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-849106560)

Need a few clarifications
> Run these tests on the simulators on which they shouldn't pass, and check that they fail indeed.
- Is this experiment meant to be done on the local machine or we need to include the tests which should fail and exclude them from validation.
- We could probably remove the failing tests in Prototype notebook after https://github.com/microsoft/qsharp-runtime/issues/323. We would need to update `AbstractKataMagic` and I haven't thought closely about it yet.

> Tests that pass on ToffoliSimulator only should allocate and manipulate a lot of qubits, like a few hundred - full state simulator should run out of memory trying to do this, though I'm not sure how it will look in notebooks, most likely a timeout.
- I tried allocating `50 qubits`( >`30 qubits` which is max limit for full state simulator) in |00..00> state and the test checks if they are succesfully converted to |11...11> state
- The Prototype notebook isn't timing out in validation when using the validation script.
- Running `dotnet test --filter Name~ToffoliSimulatorCheck` reports a time `~80 ms`, so hopefully we should be able to retain this test

Manvi-Agrawal:
Hi @tcNickolas @anjbur, I just thought of slightly cleaner approach to refractor the magics from the one mentioned [here](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-846445000).

I am thinking to get away with the generics by considering `userAnswer` as a string in both the magics.

On analysing the code, I found that `KataMagic` needed the `userAnswer` in operation format just to replace the last solution entered by the user. Hence, the process of obtaining `userAnswer`  as `OperationInfo` could be moved from `GetUserAnswer`  to a new function `FindSolution` in `KataMagic` of the following signature type :
```
protected virtual OperationInfo FindSolution(string userAnswer)
```
I would love to know your opinion on my new approach.

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
```diff
@@ -44,185 +44,76 @@ public CheckKataMagic(IOperationResolver resolver, ICompilerService compiler, IL
...
-            if (userAnswer == null) { return ExecuteStatus.Error.ToExecutionResult(); }
-
-            return Simulate(test, userAnswer, channel)
-                ? "Success!".ToExecutionResult()
-                : ExecuteStatus.Error.ToExecutionResult();
-        }
+        protected override QsNamespaceElement[] GetDeclaredCallables(string code, IChannel channel) =>
+               Compiler
+               .IdentifyElements(code)
+               .ToArray();
```
anjbur, line 64:
```suggestion
               .IdentifyElements(code);
```

anjbur (changes requested):
@Manvi-Agrawal: I agree, the generic should be able to be eliminated by moving `KataMagic`'s resolution of the operation name from `GetUserAnswer` to a `FindSolution` method that's called in `Simulate`. That should simplify `AbstractKataMagic` nicely.

From reading through the code, the changes look good so far! I'll wait to validate them more extensively until the change is made to remove the generics.

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
Manvi-Agrawal: Hi @anjbur. I unfortunately can't do this change. 

`ToArray()` method is being used to get the uniform return type from `GetDeclaredCallables(code, channel)`. <br> `GetDeclaredCallables(code, channel)` calls one of these methods according to the magics to compile or semi-compile  the code.
- **Semi-compilation in "CheckKataMagic" :** `Compiler.IdentifyElements(code)` returns `IEnumerable<QsNamespaceElement>`
- **Compilation in "KataMagic" :** `Snippets.Compile(code)` returns `QsNamespaceElement[]`

Manvi-Agrawal:
Hi @anjbur, I have changed the code to remove the generics. Unfortunately, I can't do the change [here](https://github.com/microsoft/QuantumKatas/pull/620#discussion_r663280991) due to the reason in the following [comment](https://github.com/microsoft/QuantumKatas/pull/620#discussion_r663308043).

utilities/Microsoft.Quantum.Katas/CheckKataMagic.cs
anjbur: Why not change the return type to `IEnumerable<QsNamespaceElement>` then? The result doesn't need to be an array to be used in `Compile`

Manvi-Agrawal: Hi @anjbur. Thanks for the clarification. I have now changed the return type to `IEnumerable<QsNamespaceElement>`.

Manvi-Agrawal:
Hi @anjbur. I have now changed the return type of `GetDeclaredCallables` to `IEnumerable<QsNamespaceElement>` as per your suggestion

anjbur (commented):
I left a few small changes for cleanup and to resolve build warnings, but this is looking really good!

The big thing that's left is to add testing as @tcNickolas described [here](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-849106560), through something like #621. @tcNickolas, what would be the best way to structure adding a kata for validations?

Manvi-Agrawal:
Hi @anjbur, I have applied the suggestions as requested by you.

And as for the testing changes to magics as per @tcNickolas suggestions described [here](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-849106560), I made an initial attempt in [PR#621](https://github.com/microsoft/QuantumKatas/pull/621). I have a minor confusion regarding whether the failing tests should be the part of the `Prototype` kata or not. Currently, I have excluded them from the `Prototype` kata. In my local machine, the simulator specific tests pass only on the desired simulators and fail on others. Any guidance would be truly appreciated.

tcNickolas:
I agree that testing both failing and passing tests automatically is tricky. I would still keep the failing tests in the test notebook - this way a person validating the change manually will have an easier time doing that. We can exclude the failing tests in this notebook from CI validation using the trick that you proposed https://github.com/microsoft/QuantumKatas/blob/main/.github/CONTRIBUTING.md#excluding-individual-tasks-from-validation, so that they will be only available to a manual tester.

For the sequence of merging this PR and #621:
* We'll probably want to finalize #621 in parallel with this PR, so that we can do validations in a unified way and make sure all scenarios are covered properly. 
* First we'll merge this PR.
* Then we wait for the next alpha package coming from our official builds, so that we can update the NuGet packages used in #621 with that alpha version. This is not a user-facing kata, so we can keep it on an alpha version for a bit, but I prefer not to store locally generated NuGet packages in the repo itself.
* Then we merge #621, so that validations of this change become part of our CI.
* After the next QDK release we'll switch all the packages in the repo to it, making this change available to all katas.

@anjbur How does that sound? Am I missing any validation steps we'll need?

Manvi-Agrawal:
Hi @tcNickolas thanks you giving the suggestions [here](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-878560046). I am hoping to get the changes done by the end of week. I hope that wouldn't be a problem.

tcNickolas:
Of course it's not a problem! I know it can be hard to spend time on side projects during the week - I'm juggling that myself :-)

Manvi-Agrawal:
Hi @tcNickolas , fortunately I got time to do the changes in [PR#621](https://github.com/microsoft/QuantumKatas/pull/621) as per the suggestions [here](https://github.com/microsoft/QuantumKatas/pull/621). Will try my best on resolving comments.

anjbur (commented):
I'm sorry for taking so long on this review. I hope we can get some final iteration before validating the changes and getting this in!

I've left some suggestions below. They're generally either small formatting things, adding logging, or resolving gaps in error handling. 

Additionally, after reading through the code again I think it would be good to move the `Simulate` method definition back into the `CheckKataMagic` and `KataMagic` subclasses and set it as an abstract method in `AbstractKataMagic`. While there is certainly code duplication between the two different versions of the method, there are also significant logical differences. The changes needed to unify the method increase the code complexity and currently changes behavior. For example, if a simulator logs a warning in CheckKataMagic, an exception will be thrown that will halt execution for that simulator and any others. The previous behavior would flag the warning and show a failure at the end of execution. Separating the methods should resolve any such issues with changing behavior while also simplifying the code. I should have been more clear in initially communicating what to move to a superclass versus what to leave in the subclasses, and I'm sorry for the change at this point in the process. This should put the code in a good spot between reducing duplication and simplicity/maintainability. 

Thank you for the contributions to this functionality and in 621! I'm still going through some further validations and taking a look at the Prototype kata, I'll leave comments with regards to the validations as well soon.

Manvi-Agrawal:
Hi @anjbur. Thanks a lot for your precious time while reviewing this PR. I currently have limited bandwidth, so hopefully I should be able to take a look at it in October.

utilities/Microsoft.Quantum.Katas/AbstractKataMagic.cs
```diff
@@ -0,0 +1,361 @@
...
+
+        /// <summary>
+        /// What this Magic does when triggered. It will:
+        /// - find the Test to execute based on the given name,
+        /// - compile/semi-compile the code after it found the full/partial user's answer.
+        /// - run (simulate) the test and report its result.
+        /// </summary>
+        protected virtual async Task<ExecutionResult> Run(string input, IChannel channel)
+        {
+            Assembly asm = Assembly.GetExecutingAssembly();
```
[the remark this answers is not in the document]
Manvi-Agrawal: I guess I added it for debugging purposes. I think we can remove it.

utilities/Microsoft.Quantum.Katas/AbstractKataMagic.cs
```diff
@@ -0,0 +1,361 @@
...
+            List<SimulatorBase> testSimulators = new List<SimulatorBase> ();
+
+            var testSimNames = GetSimNamesFromTestAttribute(test);
+            Logger.LogDebug($"Simulator count for {test.FullName} = {testSimNames.Count()}");
+
+            List<Assembly> simulatorAssemblies = GetSimulatorAssemblies();
+
+            foreach(var simName in testSimNames)
+            {
+                Logger.LogDebug($"Trying to create simulator of the type : {simName}");
```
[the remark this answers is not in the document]
Manvi-Agrawal: Sounds good

utilities/Microsoft.Quantum.Katas/AbstractKataMagic.cs
```diff
@@ -0,0 +1,361 @@
...
+                                testSimulators.Add(sim);
+                            }
+                            break;
+                        }
+                    }
+                    string errorMessage = $"Error while creating an instance of {simName}. " +
+                        $"{simName} not a valid execution target. " +
+                        "Either consider using a fully qualified name for the simulator. " +
+                        "or see that you aren't using Custom simulators as part of Q# projects." ;
+                    if(simulator == null)
```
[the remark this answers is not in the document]
Manvi-Agrawal: Sounds good

utilities/Microsoft.Quantum.Katas/AbstractKataMagic.cs
```diff
@@ -0,0 +1,361 @@
...
+                        if(simulator != null)
+                        {
+                            if(simulator is SimulatorBase sim)
+                            {
+                                testSimulators.Add(sim);
+                            }
+                            break;
+                        }
+                    }
+                    string errorMessage = $"Error while creating an instance of {simName}. " +
```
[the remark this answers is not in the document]
Manvi-Agrawal: Sounds  good

utilities/Microsoft.Quantum.Katas/AbstractKataMagic.cs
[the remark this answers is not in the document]
anjbur: Makes sense, removing it sounds good!

anjbur:
Hi, @Manvi-Agrawal! I wanted to check in to see if you still planned to push these changes. If not, we can make the final round of edits to get everything merged.

Manvi-Agrawal:
Hi @anjbur, apologies for keeping this PR on on hold. I think it would be good if you could do the final round of edits.

TLDR: I have done most of the changes requested by you but I am not able to validate my changes locally after updating to the qdk version using the steps mentioned in [contributing guide](https://github.com/microsoft/QuantumKatas/blob/main/.github/CONTRIBUTING.md#validating-changes-to-kata-and-check_kata-magics-on-local-machine).

  The steps in contributing guide were added by me with the guidance provided by Maria and Ryan. The steps  I were working fine until one these following qdk versions 0.17/0.18/0.19. I don't remember which version was it exactly. I think it would be good if you could update those steps for future contributors.

Thank you for your support regarding the final edits and your precious time for PR reviews.

Manvi-Agrawal:
Hi @anjbur, I finally got some time to get back on this after a really long time :-) I just tested on the latest QDK `v0.25.218240` locally that `Prototype kata` is working as per [Mariia's guidance](https://github.com/microsoft/QuantumKatas/pull/620#issuecomment-849106560)  using the steps mentioned in [contributing guide](https://github.com/microsoft/QuantumKatas/blob/main/.github/CONTRIBUTING.md#validating-changes-to-kata-and-check_kata-magics-on-local-machine). 

Are there any more changes/validations to be done? Thanks @anjbur and @tcNickolas for your guidance on this PR. Looking forward for feedback regarding `Prototype kata` introduced in https://github.com/microsoft/QuantumKatas/pull/621.

tcNickolas:
We will be migrating the Katas to the new QDK that will use a different infrastructure (see https://devblogs.microsoft.com/qsharp/introducing-the-azure-quantum-development-kit-preview/ for the announcement), including a much faster simulator, so we won't continue to improve the existing infrastructure of Q# notebooks magics. 

Thank you for your work on this, and apologies for not being able to incorporate it in our code earlier!

Manvi-Agrawal:
Thanks @tcNickolas and @anjbur for your valuable guidance. 

Looking forward to the new kata experience. Just had a look at https://quantum.microsoft.com/en-us/experience/quantum-katas, and the experience looks way more polished :-)

For my curiosity are the new katas, closed-source or open-sourced?

Manvi-Agrawal:
Thanks @tcNickolas for the link.

PR #630, opened 2021-06-10
Add qsim.DisplayDiagnostics to fix bug

### LINKED ISSUE #629: DumpOperation doesn't work inside %kata cells
A Q# Jupyter Notebook cell like this:

    %kata T_TestName
    
    operation OperationName () : Double {
        // ...
        DumpOperation(2, ApplyToFirstTwoQubitsCA(CNOT, _));
        return 0.0;
    }

does not print a matrix of a CNOT operation, as expected. Instead, it either prints nothing (if I run it on a local IQ# installation) or prints DumpMachine output for a state which I suspect is the [Choi–Jamiołkowski state](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.preparation.preparechoistatea) for the given operation (if I run it on Binder).

DumpOperation works fine for me from regular cells (executed using %simulate), so I suspect something goes wrong when redirecting output streams with `%kata` magic vs the [DumpReferenceAndTarget operation](https://github.com/microsoft/QuantumLibraries/blob/5cb1f4323e8e81196fdd2a6e07867597c62a9ef2/Standard/src/Diagnostics/Emulation/Internal.cs#L51).

QDK version: 0.17.2105143879 (both for local setup and for Binder)

@anjbur I'd really appreciate it if you could look into this!

Manvi-Agrawal:
Fixes #629.


anjbur (approved):
Looks good to me!

PR #634, opened 2021-07-03
adding QFT workbook

WaywardPhoton:
Adding a workbook for the QFT

PR #636, opened 2021-07-09
Adding QFT workbook

WaywardPhoton:
Now contains images in the workbooks, maybe the additional fix-QFT-wkbk branch can be deleted


tcNickolas (changes requested):
I see everything in place, both my changes from #634 and your new images!

However, the QFT circuit diagram is incorrect. The rotations we do are controlled, not standalone, so they have to be denoted using the controlled-gate circuit notation. The resulting diagram should look like the one in https://en.wikipedia.org/wiki/Quantum_Fourier_transform - you'll see how it first performs all controlled rotations on the first qubit to prepare it in the right state, and only then proceeds to the second qubit.
(And after that we want to show reversing the order of qubits - this is something often omitted from the QFT diagram so folks sometimes forget that it's part of the circuit.)

That's a fair amount of drawing to do, that's why I don't like circuit diagrams, especially if one has to draw them by hand :-)
I'm thinking we should get clever here and use the %trace magic to generate a circuit of the operation automatically. Here is what I got:

image

We can even do this in the notebook itself: we'll need to add another pair of cells after the code cell of task 1.7; the first one is a wrapper that allocates qubits and applies QFT to them:

```
operation DrawQFTCircuit(n : Int) : Unit {
    use qs = Qubit[n];
    QuantumFourierTransform(qs);
}
```

and the second one calling the trace command: `%trace DrawQFTCircuit n=3`.

We could  actually use a similar trick in the other task to show the circuit of swaps, to have just one circuit instead of three.

What do you think? This definitely looks like this could save you a lot of drawing effort :-)

WaywardPhoton:
Hi,

This is a great idea. I tried adding this code to the cells but it doesn't
draw out the circuit, only a block with the operation name on it.
[image: image.png]

Are you able to get this working?

tcNickolas:
Right, by default it runs an interactive diagram - you can click on the box to drill down to its components, and then click on the components, all the way down to individual gates. You can use `%config trace.defaultDepth=10` before running %trace (can be in the same cell) to set the level to which the boxes are un-boxed by default - for this circuit 10 should be enough to drill down to individual gates immediately

WaywardPhoton:
I went ahead and just re-did the image, then pushed that change to the
master branch. Hopefully it is good now :)

tcNickolas:
I think it's still incorrect... The order in which the gates are done matter, since applying rotation gates on a qubit don't commute with using that qubit as a control for another gate. So you need to do all the gates that target the first qubit first, and only then start applying Hadamard and do the gates that target the second qubit, and so on.

I feel really bad about making you redo the image manually when we have those great tools for generating the circuits from Q# code... Do you think you could try to add the cells that will generate the circuit automatically? This will have the additional benefit of the circuit being automatically correct :-)

WaywardPhoton:
I'm so sorry, I know how the QFT works and yet I managed to get the circuit
diagram wrong multiple times. I set up the %trace to work as you suggested.
Hopefully this is better now.

tcNickolas (approved):
Looks great! And no worries about not getting it right, it's like with programming itself - you can know what you're doing perfectly well and still do something that surprises even yourself :-)

Thank you so much for working on this! I'll merge as soon as the CI passes.

WaywardPhoton:
Thanks, it was great working with you!

PR #640, opened 2021-07-22
Bounded knapsack Jupyter frontend

vivanwin:


tcNickolas (approved):
Looks good! I cleaned up some formatting/wording issues and a couple of weird test names, and added `%kata` magics to the cells - most of them were missing the magic.

It looks like task 2.2 doesn't work with %kata magic - I filed #641 to track it for now and removed the code cell, same as we did in PhaseEstimation kata with the task that was covered with C# test instead of Q# test.

Thank you!

PR #643, opened 2021-07-28
Update QDK to version 0.18.2107.153439.

anjbur:


scripts/Update-QDKVersion.ps1
```diff
@@ -53,7 +53,7 @@ $dockerPath = Join-Path $katasRoot "Dockerfile"
     } | Set-Content -Path $dockerPath
 
 
-$ps1String = "Microsoft.Quantum.IQSharp --version $versionRegex"
+$ps1String = "Microsoft.Quantum.IQSharp --version 0.18.2107153439
```
tcNickolas, line 56: Please don't hardcode the new version in this script. Providing the version as a script parameter makes sure the script is reusable, and that the changes are done in all the right places. This line should be unchanged, but line 21 in this script should be updated, and the file `.github/CONTRIBUTING.md` should be updated as well.

tcNickolas (changes requested):
You can compare the version update changelist with the previous one (#633) to make sure your changes match with the ones done last time. In this case there should be one extra csproj file updated (for the new BoundedKnapsack kata), with the rest of the files matching.

Thank you!

tcNickolas (approved):
Looks good, thank you!

PR #644, opened 2021-07-30
Reduce the number of katas prebuilt on Binder

tcNickolas:
This change removes several least frequently used katas from Binder prebuild to reduce the build time on Binder.


tcNickolas:
I've checked that with this change Binder prebuild completes successfully and much faster than it used to be. The excluded katas work fine as well.

anjbur (approved):

PR #646, opened 2021-08-09
[QFT] Add workbook for part III

martinquantum:
Implementing arbitrary roots of the quantum Fourier transform (QFT) operator.


tcNickolas (approved):
Thank you!

PR #647, opened 2021-08-18
Update ReferenceImplementation.qs

lfenster:
Should take 1/2N + 1 (i.e., 2ᴺ⁻¹ + 1) steps to determine if a function is constant or balanced.  Reference implementation was actually goes through all the values (i.e, 1.. 2<<<(N-1)).

### DIFF
--- a/tutorials/ExploringDeutschJozsaAlgorithm/ReferenceImplementation.qs
+++ b/tutorials/ExploringDeutschJozsaAlgorithm/ReferenceImplementation.qs
@@ -30,7 +30,7 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
     function IsFunctionConstant_Classical_Reference (N : Int, f : (Int -> Int)) : Bool {
         // Get the first value of the function
         let firstValue = f(0);
-        // Try all the following inputs to see if any of the values differ; should be run 2ᴺ⁻¹ + 1 times
+        // Try all the following inputs to see if any of the values differ; should be run 2ᴺ⁻¹ times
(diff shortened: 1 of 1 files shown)


tutorials/ExploringDeutschJozsaAlgorithm/ReferenceImplementation.qs
```diff
@@ -30,8 +30,8 @@ namespace Quantum.Kata.DeutschJozsaAlgorithm {
     function IsFunctionConstant_Classical_Reference (N : Int, f : (Int -> Int)) : Bool {
         // Get the first value of the function
         let firstValue = f(0);
-        // Try all the following inputs to see if any of the values differ
-        for input in 1 .. 2 <<< (N - 1) {
+        // Try all the following inputs to see if any of the values differ; should be run 2ᴺ⁻¹ + 1 times
```
tcNickolas, line 33: A small fix in the comment: the loop actually runs 2ᴺ⁻¹ times - we sample 2ᴺ⁻¹+1 values but we got f(0) outside of the loop, so the loop goes through one less iteration than the total number of values we need
```suggestion
        // Try all the following inputs to see if any of the values differ; should be run 2ᴺ⁻¹ times
```

tcNickolas (changes requested):
Great catch, thank you! Just one little suggestion on the comment to make sure it matches the updated loop bounds

tcNickolas (approved):
I went ahead and applied the suggestion to the PR. Thank you!

PR #648, opened 2021-08-18
changing error rates to catch eavesdroppers

delbert:
the current error rate essentially allows any number of mis-matched qubits (ie, qubits that were eavesdropped).  lower error rate to catch eavesdropers.


tcNickolas (approved):
Wow, great catch! That sneaked in in https://github.com/microsoft/QuantumKatas/commit/3ed268465ed1b55800d95dc8541309807defbe57 which changed the meaning of threshold but didn't change the numeric value...

Thank you!!

PR #654, opened 2021-09-10
Add E91 Key Distribution kata

kurabirko:
This PR adds a new kata for E91 key distribution protocol.


KeyDistribution_E91/Tasks.qs
```diff
@@ -0,0 +1,178 @@
...
+
+    //////////////////////////////////////////////////////////////////
+    // Part I. Preparation
+    //////////////////////////////////////////////////////////////////
+	
+    // Task 1.1. Entangled Pairs
+    // Inputs:
+    //      1) "qsAlice": an array of N qubits in the |0⟩ states,
+    //	    2) "qsBob": an array of N qubits in the |0⟩ states,
+    // Goal: Create entangled pairs |Ψ⟩ = 1/√2 (|00⟩ + |11⟩) out of corresponding qubits of
```
tcNickolas, line 25: Which protocol description are you following? The one I find clearest is https://ux1.eiu.edu/~nilic/Nina's-article.pdf, and it suggests that the pair should be 1/√2 (|01⟩ - |10⟩), which then should yield anti-correlated results for matching bases choice, unlike in BB84 protocol.

KeyDistribution_E91/Tasks.qs
```diff
@@ -0,0 +1,178 @@
...
+        let bases = new Int[0];
+        // ...
+
+        return (bases, results);
+
+    }
+
+    // Task 2.3 Generate the shared key
+    // After both Alice and Bob make their measurements and share their choices of bases, it
+    // is possible to select compatible measurements and create a key out of these. You
```
tcNickolas, line 77: This should probably have a note about which measurement bases are compatible

KeyDistribution_E91/Tasks.qs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // Task 2.3 Generate the shared key
+    // After both Alice and Bob make their measurements and share their choices of bases, it
+    // is possible to select compatible measurements and create a key out of these. You 
+    // should compute that key for either Alice or Bob.
+    // Inputs:
+    //     1) "basesAlice": Alice's measurement bases multipliers,
+    //     2) "basesBob": Bob's measurement bases multipliers,
+    //     3) "results": Either Alice's or Bob's measurement outcomes.
```
tcNickolas, line 82: If I understand correctly, Alice's and Bob's results with matching bases are going to be different; in this case we should say explicitly which ones we provide as a result. Might add a Bool parameter `alicesResults` which is true if results are from Alice and false if they are from Bob, so that the solution needs to be careful about interpreting the results.

KeyDistribution_E91/ReferenceImplementation.qs
```diff
@@ -0,0 +1,199 @@
...
+        return (bases, results);
+
+    }
+
+    // Task 2.3 Generate the shared key
+    function GenerateSharedKey_Reference (basesAlice: Int[], basesBob: Int[], results: Result[]) : Bool[] {
+        mutable key = new Bool[0];
+
+        for (a, b, r) in Zipped3(basesAlice, basesBob, results) {
+            if a == b {
```
tcNickolas, line 77: This is a placeholder solution, right? There should be cases in which Alice's and Bob's bases are different but they can still use the results because the bases are compatible.

KeyDistribution_E91/ReferenceImplementation.qs
```diff
@@ -0,0 +1,199 @@
...
+
+        return s;
+    }
+
+
+    // Task 2.5 Putting it all together 
+    @Test("QuantumSimulator")
+    operation T25_E91Protocol_Reference () : Unit {
+        // 1. Alice and Bob choose basis multipliers
+        let basisListAlice = [0, 1, 2];
```
tcNickolas, line 123: Should this be hardcoded property of the protocol rather than the participant's choice? I interpreted the choice of basis multipliers step in the task description as choosing the random base for each qubit, rather than choosing the bases available to them

KeyDistribution_E91/Tests.qs
```diff
@@ -0,0 +1,102 @@
...
+                AssertQubitWithinTolerance(r, q, 1e-5);
+            }
+            Reset(q);
+        }
+
+    }
+
+    @Test("QuantumSimulator")
+    operation T23_GenerateSharedKey() : Unit {
+        // ...
```
tcNickolas, line 88: This test should be similar to task 2.4 from BB84 kata, purely classical logic

KeyDistribution_E91/ReferenceImplementation.qs
```diff
@@ -0,0 +1,199 @@
...
+            -  expectationValues[1]
+            + expectationValues[2] 
+            + expectationValues[3]; 
+
+        return s;
+    }
+
+
+    // Task 2.5 Putting it all together 
+    @Test("QuantumSimulator")
```
tcNickolas, line 120: The reference shouldn't be marked as a test - if it's marked it shows up with the rest of the tests, and it can be confusing for the learner (same for T32_E91ProtocolWithEavesdropper_Reference)

tcNickolas (changes requested):
I made the first pass and left some comments before our discussion. Thank you!

PR #655, opened 2021-09-10
Add weak coloring

### LINKED ISSUE #139: Explore more graph coloring problems
The latest [GraphColoring](https://github.com/microsoft/QuantumKatas/tree/master/GraphColoring) kata covers one type of graph coloring problems - vertex coloring (assigning colors to graph vertices so that no two vertices connected with an edge are assigned the same color).

There exists a variety of other graph coloring problems (see [Wikipedia](https://en.wikipedia.org/wiki/Graph_coloring#Other_colorings) for a list). For some of them the problem encoding can be made small enough to make it possible to simulate the Grover's algorithm and to solve small but non-trivial problem instances. It would be great to add chapters on solving these problems. Here are several that look the most promising:

* [Weak coloring](https://en.wikipedia.org/wiki/Weak_coloring) - vertex coloring where every non-isolated vertex has at least one neighbor with a different color.
* [Sum coloring](https://en.wikipedia.org/wiki/Sum_coloring) - vertex coloring that minimizes the sum of colors used on vertices. This opens the way for using Grover's algorithm for solving minimization problems, which is a separate interesting topic.

Let me know if you want to try and implement one of these topics!

Manvi-Agrawal:
Resolves #139


GraphColoring/Tasks.qs
```diff
@@ -149,4 +150,109 @@ namespace Quantum.Kata.GraphColoring {
...
+    // Inputs:
+    //      1) The number of vertices in the graph V (V Γëñ 6).
+    //      2) An array of E tuples of integers, representing the edges of the graph (E Γëñ 12).
+    //         Each tuple gives the indices of the start and the end vertices of the edge.
+    //         The vertices are indexed 0 through V - 1.
+    //      3) An array of 2V qubits colorsRegister that encodes the color assignments.
+    //      4) A qubit in an arbitrary state |yΓƒ⌐ (target qubit).
+    //      5) A vertex in the graph, indexed 0 through V - 1
+    //
+    // Goal: Transform state |x, yΓƒ⌐ into state |x, y Γèò f(x)Γƒ⌐ (Γèò is addition modulo 2),
```
tcNickolas, line 218:
```suggestion
    // Goal: Transform state |x, y⟩ into state |x, y ⊕ f(x)⟩ (⊕ is addition modulo 2),
```

GraphColoring/Tests.qs
```diff
@@ -267,4 +272,103 @@ namespace Quantum.Kata.GraphColoring {
...
+    @Test("QuantumSimulator")
+    operation T322_IsWeakColoringValid () : Unit {
+        let testCases = ExampleGraphs();
+        let exampleColoringForThreeVertices = [[0, 0, 0], [2, 1, 3]];
+        let exampleColoringForFourVertices = [[0, 2, 1, 3], [3, 2, 0, 0], [0, 2, 1, 0] ];
+        let exampleColoringForFiveVertices = [[0, 1, 2, 3, 1], [0, 1, 1, 2, 2], [0, 0, 1, 2, 3]];
+
+
+        let (V0, edges0) = testCases[0];
+        for (coloring, expectedResult) in Zipped(exampleColoringForThreeVertices, [true, true]) {
```
tcNickolas, line 319: Nice use of Zipped to avoid repeating facts for each graph! I think you can make it even nicer if you create an array of colorings and expected verdicts for each graph and zip that array with ExampleGraphs(), then just iterate over that big array checking each entry.

So something along the lines of 
```qsharp
let coloringsAndVerdicts3 = Zipped(exampleColoringForThreeVertices, [true, true]);
...
let fullTestCases = Zipped(testCases, [coloringsAndVerdicts3, coloringsAndVerdicts4, ...]);
```

GraphColoring/Tests.qs
```diff
@@ -267,4 +272,103 @@ namespace Quantum.Kata.GraphColoring {
...
+            Fact(IsVertexWeaklyColored(V_t, edges_t, colorings[t], vertex) == expectedResults[t][vertex],
+                $"Vertex = {vertex} judged {not expectedResults[t][vertex]} for coloring = {colorings[t]} , edges = {edges_t}");
+            }
+        }
+    }
+
+    @Test("QuantumSimulator")
+    operation T322_IsWeakColoringValid () : Unit {
+        let testCases = ExampleGraphs();
+        let exampleColoringForThreeVertices = [[0, 0, 0], [2, 1, 3]];
```
tcNickolas, line 313: Are there any test cases that are interesting for the weak graph coloring but not for the regular graph coloring, both in terms of graph structure and its coloring? It makes sense to add notes on the way you selected the colorings for the graphs to see if any corner cases are missing, similar to the comments on the way we selected example graphs themselves.

tcNickolas (changes requested):
I made a high-level pass and left some comments. I focused on the tasks, reference solutions and tests, didn't look at the notebooks - I'll review those once we're happy with the core part and you update them to match the changes in the .qs files.

One more note, could you use American spelling (color, neighbor) consistently, rather than alternating between that and British? Inconsistency in such things

Thank you for working on this!

Manvi-Agrawal:
Hi @tcNickolas, I am still doing requested changes. Will inform you when it's ready to review.

GraphColoring/ReferenceImplementation.qs
```diff
@@ -169,4 +169,96 @@ namespace Quantum.Kata.GraphColoring {
...
+            if colors[start] != colors[end] {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    // Task 3.2.2 Classical verification of weak coloring
+    function IsWeakColoringValid_Reference (V : Int, edges: (Int, Int)[], colors: Int[]) : Bool {
+        mutable validBits = new Bool[V];
```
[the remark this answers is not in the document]
Manvi-Agrawal: I realised this while writing code in workbook. I have also updated the reference solution now

GraphColoring/Tests.qs
Manvi-Agrawal: Done. Thanks for this suggestion. Made the code lot more cleaner

GraphColoring/Tasks.qs
```diff
@@ -149,4 +150,109 @@ namespace Quantum.Kata.GraphColoring {
...
+    //      1) The number of vertices in the graph V (V Γëñ 6).
+    //      2) An array of E tuples of integers, representing the edges of the graph (E Γëñ 12).
+    //         Each tuple gives the indices of the start and the end vertices of the edge.
+    //         The vertices are indexed 0 through V - 1.
+    //      3) An array of V integers, representing the vertex coloring of the graph.
+    //         i-th element of the array is the color of the vertex number i.
+    //      4) A vertex in the graph, indexed 0 through V - 1
+    // Output: true if the vertex is weakly colored
+    //         (i.e atleast one neighbouring vertex of different color),
+    //         and false otherwise.
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

GraphColoring/Tests.qs
Manvi-Agrawal: Hi added a note for the same. I just explored the current graphs from the Coloring perspective. Using trivial cases to make things simple. The interesting part about these graphs is that since graph 5 is a superset of graph 3 and 4, any weak coloring that passes graph 3 or 4, would also pass graph 5

GraphColoring/Tasks.qs
```diff
@@ -149,4 +150,109 @@ namespace Quantum.Kata.GraphColoring {
         // ...
         return new Int[V];
     }
+
+    //////////////////////////////////////////////////////////////////
+    // Part III. Weak coloring problem
+    //////////////////////////////////////////////////////////////////
+
+    // Task 3.1. Determine if an edge starts with mentioned vertex
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

GraphColoring/Tasks.qs
```diff
@@ -149,4 +150,109 @@ namespace Quantum.Kata.GraphColoring {
...
+    // Output: true if the edge starts with vertex provided
+    //         and false otherwise.
+    // Examples: edge (0,1) starts with vertex 0, so return true
+    //           edge (0,1) does not start with vertex 1, so return false
+    function DoesEdgeStartWithVertex(edge: (Int, Int), vertex : Int) : Bool {
+        // ...
+        return false;
+    }
+
+    // Task 3.2.1 Determine if a vertex is weakly colored
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

Manvi-Agrawal:
Hi @tcNickolas, I have made the desired changes. Could you please take a look at my changes and the Jupyter notebooks. Thanks

Manvi-Agrawal:
Hi @tcNickolas. It's weird that `T35_WeakColoringVertexOracle` is facing timeout exception in Windows platform. I write code on windows and this test passed reliably on my local machine. Running it with dotnet gives a time about 2 minutes which should be okay perhaps since we have a timeout after 5 minutes.

Manvi-Agrawal:
Otherwise I can maybe update this test case to not iterate over all the colorings or maybe exclude  more graphs. Thoughts?

tcNickolas:
Hmm, let me try re-running that job real quick to see if this timeout persists or if it was transient - for example, sometimes we get slower machine in CI so overall job run takes longer

Manvi-Agrawal:
Thank you @tcNickolas

tcNickolas:
The re-run failed as well, so doesn't look like an accident... Let me do another round of review, and I'll let you know what I think

Manvi-Agrawal:
Sounds good

tcNickolas (approved):
Apologies again for the delay reviewing this PR! I did the normal cleanup, didn't find any major issues.

For task 3.5, the test's runtime depends on the machine on which it runs - I ran the test on my laptop with just the first four graphs from ExampleGraphs (even without the 5th one that you included), and it took almost 5 minutes! I can definitely see how five graphs could take over 5 minutes on CI machines. The first three graphs seem to take a reasonable amount of time, though, just over a minute, so I'll try that. We'll see if that passes CI - fingers crossed!

Thank you so much for working on this!

PR #665, opened 2021-09-30
Added entanglement swapping as an extra task to the Teleportation kata

filipw:
Hello 👋 I thought it might be an interesting task to have entanglement swapping as a task in the Teleportation kata.

The user starts with two maximally entangled `Phi^{+}` pair, one belonging to Alice and the other to Bob. The task is to use one qubit of each pair to produce an entanglement of the other two qubits - without any interaction between those.

### DIFF
--- a/Teleportation/ReferenceImplementation.qs
+++ b/Teleportation/ReferenceImplementation.qs
@@ -75,12 +75,12 @@ namespace Quantum.Kata.Teleportation {
     }
     
     // Task 1.8. Entanglement swapping
-    operation EntanglementSwapping_Reference (qubitPairAlice : Qubit[], qubitPairBob : Qubit[]) : Unit {
-        CNOT(qubitPairAlice[0], qubitPairBob[0]);
-        H(qubitPairAlice[0]);
+    operation EntanglementSwapping_Reference (entangledPairAlice : Qubit[], entangledPairBob : Qubit[]) : Unit {
+        CNOT(entangledPairAlice[0], entangledPairBob[0]);
+        H(entangledPairAlice[0]);
         
-        let classicalBits = (M(qubitPairAlice[0]) == One, M(qubitPairBob[0]) == One);
-        ReconstructMessage_Reference(qubitPairBob[1], classicalBits);
+        let classicalBits = (M(entangledPairAlice[0]) == One, M(entangledPairBob[0]) == One);
+        ReconstructMessage_Reference(entangledPairBob[1], classicalBits);
     }
     
     //////////////////////////////////////////////////////////////////
--- a/Teleportation/ReferenceImplementation.qs
+++ b/Teleportation/ReferenceImplementation.qs
@@ -74,15 +74,13 @@ namespace Quantum.Kata.Teleportation {
         return Measure([basis], [qBob]) == One;
     }
     
-    // Task 1.7. Entanglement swapping
-    operation EntanglementSwapping_Reference (qubitPairAlice : Qubit[], qubitPairBob : Qubit[]) : (Bool, Bool) {
+    // Task 1.8. Entanglement swapping
+    operation EntanglementSwapping_Reference (qubitPairAlice : Qubit[], qubitPairBob : Qubit[]) : Unit {
         CNOT(qubitPairAlice[0], qubitPairBob[0]);
         H(qubitPairAlice[0]);
         
         let classicalBits = (M(qubitPairAlice[0]) == One, M(qubitPairBob[0]) == One);
         ReconstructMessage_Reference(qubitPairBob[1], classicalBits);
-
-        return (M(qubitPairAlice[1]) == One, M(qubitPairBob[1]) == One);
     }
     
     //////////////////////////////////////////////////////////////////
--- a/Teleportation/ReferenceImplementation.qs
+++ b/Teleportation/ReferenceImplementation.qs
@@ -74,6 +74,16 @@ namespace Quantum.Kata.Teleportation {
         return Measure([basis], [qBob]) == One;
     }
     
+    // Task 1.7. Entanglement swapping
+    operation EntanglementSwapping_Reference (qubitPairAlice : Qubit[], qubitPairBob : Qubit[]) : (Bool, Bool) {
+        CNOT(qubitPairAlice[0], qubitPairBob[0]);
+        H(qubitPairAlice[0]);
+        
+        let classicalBits = (M(qubitPairAlice[0]) == One, M(qubitPairBob[0]) == One);
+        ReconstructMessage_Reference(qubitPairBob[1], classicalBits);
+
+        return (M(qubitPairAlice[1]) == One, M(qubitPairBob[1]) == One);
+    }
     
     //////////////////////////////////////////////////////////////////
     // Part II. Teleportation using different entangled pair
@@ -177,5 +187,4 @@ namespace Quantum.Kata.Teleportation {
             X(qCharlie);
         }
     }
-    
 }
--- a/Teleportation/Tasks.qs
+++ b/Teleportation/Tasks.qs
@@ -170,7 +170,7 @@ namespace Quantum.Kata.Teleportation {
     // The first pair belongs to Alice and the second to Bob. 
     //
     // Hint: You may find your answers for 1.2 and 1.3 useful, as similar steps are needed here
-    operation EntanglementSwapping (qubitPairAlice : Qubit[], qubitPairBob : Qubit[]) : Unit {
+    operation EntanglementSwapping (entangledPairAlice : Qubit[], entangledPairBob : Qubit[]) : Unit {
         // ...
     }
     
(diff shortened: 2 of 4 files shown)


Teleportation/ReferenceImplementation.qs
```diff
@@ -74,6 +74,14 @@ namespace Quantum.Kata.Teleportation {
         return Measure([basis], [qBob]) == One;
     }
     
+    // Task 1.8. Entanglement swapping
+    operation EntanglementSwapping_Reference (entangledPairAlice : Qubit[], entangledPairBob : Qubit[]) : Unit {
+        CNOT(entangledPairAlice[0], entangledPairBob[0]);
```
tcNickolas, line 79: Can we use SendMessage_Reference here instead of the individual gates to emphasize that it's the same protocol?

Teleportation/Tasks.qs
```diff
@@ -150,6 +150,29 @@ namespace Quantum.Kata.Teleportation {
...
+    //
+    // Charlie can now teleport the state of Alice's qubit onto Bob's qubit, thus teleporting the entanglement. 
+    // Just like in "standard" teleportation, Bob still needs to apply the reconstruction steps -
+    // based on Charlie's measurement results - to the qubit still in his possession.
+    //
+    // The practical outcome is that the entangled state |Φ⁺⟩ = (|00⟩ + |11⟩) / sqrt(2) now spans across 
+    // Alice's and Bob's qubits - the ones that they didn't send to Charlie. They are now maximally entangled
+    // even though they never interacted in the first place!
+    //
+    // Goal: Entangle one part of Alice's input qubits pair with one part of Bob's input qubit pair,
```
tcNickolas, line 166: The way the task is formulated now, it's not clear which of Alice's qubits should end up entangled with which of Bob's. What is worse, the way the test is written, there is nothing preventing the solution from just swapping entangledPairAlice[0] with entangledPairBob[0], yielding not one but two pairs of entangled qubits. And it's pretty hard to verify in a test that two specific qubits never interact in a single multi-qubit gate... (This is exactly the same problem task 1.4 has, actually. We have tasks in which we prohibit all multi-qubit gates other than measurements, but that's slightly different)

This task is really about figuring out which of the building blocks of teleportation to apply at which location, since it doesn't use any new blocks. I'd make this task an open-ended one, similar to task 1.7, so that the learner has to figure the steps of the protocol, prepare the pairs, apply the right sequence of the previous tasks and figure out how to demonstrate it - but it's up to them to figure out which qubits they want to entangle in the end, and we don't have to develop a test to validate that the solution indeed performs the teleport :-)

tcNickolas (changes requested):
Ah, it's a challenging task to do kata-style, I've been thinking about it for a while but couldn't come up with a good way. I've left some comments about my concerns for this task, let's see if we can finally make it :-) Thank you!

Teleportation/ReferenceImplementation.qs
filipw: yes - changed that, thanks!

Teleportation/Tasks.qs
filipw: Thanks a lot for a quick and very helpful feedback. I now see your point of view when approaching these katas better.

I have reworked the kata to incorporate your suggestions. With the updated version, the reference implementation is only 2 lines, which is nice because all it boils down to is to reuse previous task solutions in a different context. Also with  the update version, the user no longer gets both EPR pairs but instead both Bob's qubits but only one part of the pair from Alice.

The task is then rephrased to use teleportation to create entanglement between Bob's 2nd qubit and Alice's 2nd qubit to - which the user is not even given access to. The only place that accesses Alice's 2nd qubit is the test, which still verifies the result. I hope this is now conveying the intent and the expectations for the user better and that it allows us to keep the test for feedback purposes. 

However, if you wish, I can rework this into an open-ended task (where everything, including the initial entanglement of both pairs, has to be done by the user) as well!

tcNickolas: As long as the code has access to one of the qubits from Alice's pair and both qubits from Bob's pair, the SWAP solution still passes - swap the qubit from Alice's pair with the one from Bob's pair that is supposed to be part of the final entangled state. To prevent this from happening, we need to give the code access to Charlie's qubits only, which becomes effectively task 1.2 again...

I wonder if we can implement this task in a more functional programming style - not give access to the qubits directly, but rather have a task which takes no arguments and returns two operations - one that Charlie should do (takes 2 qubits as inputs and returns an integer) and one that Bob should do (takes a qubit and an integer). This way we never give the learner's code access to the qubits they shouldn't have, so it doesn't have a way to do SWAP. The testing harness then would call these methods sequentially and check that as a result Alice's and Bob's qubits are in the |00>+|11> state. And the reference solution would combine reference solutions to tasks 1.2 and 1.3 with a wrapper to convert a tuple of Boolean values to a single integer and vice versa (or the learner can try and come up with something different).

What do you think about this approach?

filipw: of course you are right 🤦‍♂️

I have changed the task to the functional style as suggested - it's a great idea. The goal is now to produce a pair of callables to be returned in a tuple, as suggested. The rest is pretty much the same as before.

Teleportation/Tests.qs
```diff
@@ -189,6 +189,28 @@ namespace Quantum.Kata.Teleportation {
         TeleportPreparedStateTestLoop(PrepareAndSendMessage_Reference, ReconstructAndMeasureMessage);
     }
     
+    @Test("QuantumSimulator")
+    operation T18_EntanglementSwapping () : Unit {
+        
+        let numRepetitions = 100;
```
tcNickolas, line 195: You're using this variable in just one place, so probably don't need a separate variable for it. (Also, if the loop starts with 0, the actual number of repetitions is one greater than the value of the variable)

Teleportation/Tests.qs
```diff
@@ -189,6 +189,28 @@ namespace Quantum.Kata.Teleportation {
...
+
+        for i in 0 .. numRepetitions {
+            use (qAlice1, qAlice2) = (Qubit(), Qubit());
+            H(qAlice1);
```
tcNickolas, line 199: You can use Entangle_Reference here and below to make the code shorter and more self-documenting

Teleportation/Tests.qs
```diff
@@ -189,6 +189,28 @@ namespace Quantum.Kata.Teleportation {
...
+
+            use (qBob1, qBob2) = (Qubit(), Qubit());
+            H(qBob1);
+            CNOT(qBob1, qBob2);
+            
+            let (teleportOp, adjustOp) = EntanglementSwapping();
+            let result = teleportOp(qAlice1, qBob1);
+            adjustOp(qBob2, result);
+
+            let (c1, c2) = (M(qAlice2) == One, M(qBob2) == One);
```
tcNickolas, line 210: This test is a bit too relaxed to verify that the resulting state of qubits qAlice2 and qBob2 is (|00⟩ + |11⟩) / sqrt(2); it only checks that measurement results match, which would also be the case for the state (|00⟩ - |11⟩) / sqrt(2), or any other superposition of the basis states |00⟩ and |11⟩.

A better way to verify that the state of the qubits is the state you'd expect is the approach used in the Superposition kata (detailed in [this blog post](https://devblogs.microsoft.com/qsharp/inside-the-quantum-katas-part-1/)): apply adjoint of the operation that would prepare these qubits in the required state, and use `AssertAllZero` to check that the state of the qubits afterwards is |0...0⟩.

This will also allow to reduce the number of loop iterations: since now we only need to cover the 4 possible outcomes of the measurement in teleportation process, rather than to gather statistics on the measurement outcomes for a superposition state, we can get away with something like 10-15 iterations instead of a hundred.

Teleportation/Teleportation.ipynb
```diff
@@ -223,34 +222,79 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "source": [
+    "%kata T18_EntanglementSwapping\n",
+    "\n",
+    "operation EntanglementSwapping () : ((Qubit, Qubit) => Int, (Qubit, Int) => Unit) {\n",
+    "    // ...\n",
+    "    fail \"Not implemented yet\";\n",
```
tcNickolas, line 290: We don't need a fail in the notebook version of the task, since the notebook cells don't have to compile successfully for the previous ones to work, unlike the Q# projects

tcNickolas (changes requested):
I like how this task is shaping! Just a few more comments on the test implementation, to make sure we're testing the right thing.

Thank you!

Teleportation/Tests.qs
filipw: done

Teleportation/Tests.qs
filipw: I should have done this from the beginning - this is obviously a much better way - done

Teleportation/Tests.qs
filipw: done

filipw:
> and fixed the test - it was using the reference implementation of the task instead of the learner's solution.

oops was a left over from... well, testing 😅
thanks a lot @tcNickolas!

PR #673, opened 2021-10-15
Workbook for TruthTables

olstulo:
Added a 1st draft of a workbook for TruthTables


tcNickolas (approved):
Looks good! I went ahead and added the per-task links between the kata and the workbook, and the code cells in the workbook that show the complete solutions. I also wrote up a bit more details for the tasks that needed them - most katas don't use user-defined types, so I imagine the syntax for dealing with them will be new to a lot of users. I also removed the last two tasks which don't have solutions - not having a solution explained in the workbook is enough, we don't need to put additional pressure on ourselves by saying "coming soon" :-)

Thank you! Let me know if you'd like to continue working on the solutions to the last two tasks!

olstulo:
> Looks good! I went ahead and added the per-task links between the kata and the workbook, and the code cells in the workbook that show the complete solutions. I also wrote up a bit more details for the tasks that needed them - most katas don't use user-defined types, so I imagine the syntax for dealing with them will be new to a lot of users. I also removed the last two tasks which don't have solutions - not having a solution explained in the workbook is enough, we don't need to put additional pressure on ourselves by saying "coming soon" :-)
> 
> Thank you! Let me know if you'd like to continue working on the solutions to the last two tasks!

Thank you for taking the time to review my PR and making it merge ready! And yes, I'd like to continue working on the last 2 solutions.

tcNickolas:
Great, thank you!!

Issue #542, opened 2020-10-22
 Create workbooks for the katas/tutorials, part 2


tcNickolas:
(This issue is the second part of #208)

The katas offer a set of programming exercises on quantum computing and reference solutions to them, but we often hear that it would be helpful to have the solutions explained, with the logic steps necessary to arrive from the problem description to the code spelled out. It is especially important for people who go through the katas on their own, without a study group to support them.

To address this, we came up with the idea of "workbooks" - explanations of the solutions to the tasks offered in the katas/tutorials, written in Jupyter Notebook format. You can check any of the workbooks completed in #208 to get an idea of what the workbooks look like.

Here are the katas that don't have a workbook yet and seem like a good fit for one.

* [ ] PhaseEstimation
* [ ] QFT
* [ ] RippleCarryAdder
* [ ] TruthTables

born-2learn:
I would love to work on them, @tcNickolas. Will it be okay if I move with a slow pace for these?

tcNickolas:
@born-2learn Sure! Usually I recommend to pick one workbook to work on and claim it, so that if somebody else wants to take another one, they can; once that one is complete, can pick another one if you want, and so on. The #208 took about a year to complete :-) (though it had more workbooks in it)

born-2learn:
@tcNickolas Alright, I would like to try out the Phase Estimation workbook then.

WaywardPhoton:
@tcNickolas  Is there anything I can work on? I'd be happy to create workbooks for QFT if that is something that would be needed.  It seems like there isn't one currently available.

WaywardPhoton:
@tcNickolas  Is there anything I can work on? I'd be happy to create workbooks for QFT if that is something that would be needed.  It seems like there isn't one currently available.

tcNickolas:
Hi Ana,
Absolutely! I'm not aware of anybody else working on the QFT kata workbook,
so you're welcome to it! Please let me know if you need any help from me to
get started.

WaywardPhoton:
@tcNickolas  I'd love some general guidelines on content. I know that the solution already exists (and all I have to do is explain it) but if there is anything specific that explanations are required to have I'd like to know!

tcNickolas:
The best way to get an idea of how the workbook should look is probably
checking out the ones for other katas. Typically we want to explain the way
to come to the solution mathematically, and then to highlight some Q#
features/library operations that come in handy but can be not obvious for
someone who is just getting started with the language.

[...]

PR #674, opened 2021-10-16
[Workbook Added] Solutions provided for Task 1.1 and 1.2

devikamehra:
As per the Issue #542, I have started with Key_Distribution_BB84 Workbook. So far I have covered Task 1.1 and Task 1.2 only with an intention to check if I am going in the right direction. There are few hyperlinks added in the Key_Distribution_BB84 quantum kata as well to ensure the Workbook works smoothly.


tcNickolas (approved):
Looks great! I did a couple of small edits to polish it a bit, didn't make sense to send the PR back for the small things.

I'll merge this PR once the CI build passes, and then you can open a separate one if you want to keep working on this workbook.
Thank you!

PR #677, opened 2021-10-17
Try saving file with jupyter nb instead of jup lab

TonyHoldroyd:
Try saving file with jupyter nb instead of jup lab


TonyHoldroyd:
Just checking this for conflicts

TonyHoldroyd:
I switched to Jupyter Notebooks on my  laptop that has a fresh install of everything Q# and finally got a big green tick :)). Hope it passes muster, for Tasks 5 -7 but if not I will do some more work on it now I know it's finally uncorrupted and actual JSON.
Jupyter lab should come with a public health warning as far as github goes.
BTW, github was corrupting any cells that had more in them than pure $$ ...$$ equations. Might have been down to gitlab as well, but my equation cells are all pure equation.

TonyHoldroyd:
Not sure if I need to reopen this

TonyHoldroyd:
Yes please Mariia. I'll close the others. I uses Jupyter Notebooks, which I
should have done all along

PR #689, opened 2021-11-17
[CounterSimulator] Return to counting callables rather than strings

tcNickolas:
The newly added test shows the scenario in which counting strings fails: the same operation called as part of different partial applications with different parameters. Counting callables works, since each partial application produces a different callable.


tcNickolas:
Closing this PR, since the change it was supposed to fix was reverted. Will be incorporated in the next iteration

PR #690, opened 2021-11-17
Adding steps to use current built artifacts to validate Katas

anpaz:


tcNickolas (changes requested):
The CI build failure we're getting on this PR doesn't look right - the issue was limited to CounterSimulator so it should only affect the notebooks which use it, and with this change the CI build fails on the first cell of each notebook, including the ones which use only QuantumSimulator. Could it be that the CI step cannot find M.Q.Katas package at all when validating the notebooks?

tcNickolas:
One more thought: since the end users are always running the Katas with the released QDK, we need to keep some of the validation jobs on the latest released version, otherwise we'll risk introducing changes to the katas that use not-yet-released functionality. 

Maybe we should keep the existing jobs as is, testing the content of all notebooks, and add a separate one which would test a subset of notebooks (the ones which use all the features of CounterSimulator) on the current version of the M.Q.Katas package, testing that functionality specifically?

tcNickolas:
I filed https://github.com/microsoft/QuantumKatas/issues/742 to keep track and the context of this issue. I'll close this PR, since it's not in active development and we'll probably need a different approach, but will keep the branch.

PR #705, opened 2021-12-12
Update SingleQubitGates.ipynb Description of Pauli Matrices

lwoodyiii:
Changed Description of Pauli Matrices from just Self-Adjoint to Hermitian AND Unitary. A matrix has to be Hermitian AND Unitary to be equal to its own inverse. Being Hermitian is not enough.
I acknowledge Self-Adjoint and Hermitian are used interchangeably (I just prefer Hermitian :) )

### DIFF
--- a/tutorials/SingleQubitGates/SingleQubitGates.ipynb
+++ b/tutorials/SingleQubitGates/SingleQubitGates.ipynb
@@ -259,7 +259,7 @@
    "source": [
     "## Pauli Gates\n",
     "\n",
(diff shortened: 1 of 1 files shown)


tcNickolas:
I'm not sure this is a change I want to make. 
* For unitarity of Pauli gates, we say explicitly in the beginning of the tutorial that all gates are unitary, so Pauli gates have no choice but to be unitary :-) 
* After that, Hermitian and self-adjoint can be used interchangeably, and I prefer self-adjoint because this term is self-descriptive. If the tutorial says something is self-adjoint, the reader doesn't need to do the extra effort of recalling what Hermitian means - they can figure out right away that it's something that is adjoint to itself. A lot of our users don't have extensive practice with math terminology, so the fewer mental steps we make them do, the better

lwoodyiii:
sounds good, np

PR #711, opened 2021-12-17
Added Basic Gates Workbook Tasks 1.9, 1.10

TonyHoldroyd:
Added Basic Gates Workbook Tasks 1.9, 1.10

### DIFF
--- a/BasicGates/Workbook_BasicGates.ipynb
+++ b/BasicGates/Workbook_BasicGates.ipynb
@@ -2,6 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "7857a5f0",
    "metadata": {
     "tags": []
    },
@@ -19,6 +20,7 @@
   },
   {
    "cell_type": "markdown",
+   "id": "00738ea4",
    "metadata": {},
    "source": [
     "**What you should know for this workbook**\n",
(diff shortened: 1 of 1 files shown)


BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1017,10 +1091,667 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "24ef1b32",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%kata T110_BellStateChange3\n",
```
tcNickolas, line 1729: I think the CI build is not happy about this code cell: for the workbooks, it checks that the code provided in the cell passes the test (as opposed to the main kata notebooks, in which the CI build replaces %kata magic with %check_kata magic which, in turn, checks that the reference solution to this task is correct).
To achieve the same pedagogical effect without upsetting the CI build, you'll need to say something like "try to add lines `Z(qs[0]); X(qs[0]);`" to the previous cell and see how that changes the input state

TonyHoldroyd:
Ciao Maria, thank you for looking at this, I didn't realise that you saw
draft pull requests, I'll give your suggestion a go. At least there aren't
any weird IDs this time :)

tcNickolas:
I see all pull requests, if I look :-) I don't look closely into drafts, but for this one the CI build was failing and I got wondering why it was failing.

TonyHoldroyd:
Okay, thanks, I expect it's better to go with draft pull requests though
when I'm unsure of successful parsing

PR #717, opened 2021-12-20
Update QDK to .NET6.0 (QuantumKatas)

ricardo-espinoza:
We're migrating the QDK to the most recent Long Time Support version of the .NET framework. For details about this change, refer to the original issue [Q#C:1224](https://github.com/microsoft/qsharp-compiler/issues/1224).

As part of this change, we're:

- Re-targeting all .NetCoreApp3.1 binaries to .NET6.0
- Updating Docker images, samples and templates.
- Libraries using .NetStandard2.1 are not affected by this change.
- The minimum supported .NET version in the QDK will also be updated from 3.1 to 6.0

This is a multi-repo change, and all changes will need to be merged concurrently.


tcNickolas:
Can you please convert this PR to a draft? This will help emphasize that it will be merged only after the next release

tcNickolas:
The CI builds are failing because of the new tutorial MultiQubitSystemMeasurements, its csproj file needs to be updated as well

PR #720, opened 2022-01-02
testing task 1.10 v3 with draft PR

TonyHoldroyd:


TonyHoldroyd:
Hello Mariia, my latest pR failed validation with all of the part 2 notebook tests with the same error: fail: Microsoft.Quantum.IQSharp.References[0]
      Unable to load package 'Microsoft.Quantum.Standard.Visualization':  Unable to find package 'Microsoft.Quantum.Standard.Visualization'. Existing packages must be restored before performing an install or update.
Any ideas please?
Thanks

TonyHoldroyd:
All the failiure appear to be:
fail: Microsoft.Quantum.IQSharp.References[0]
      Unable to load package 'Microsoft.Quantum.Standard.Visualization':  Unable to find package 'Microsoft.Quantum.Standard.Visualization'. Existing packages must be restored before performing an install or update.

tcNickolas:
The "Microsoft.Quantum.Standard.Visualization" message is just noise, it doesn't affect the test passing or failing - you'll notice it's reported on all notebooks regardless of their validation status

This failure is specific to the cell that solves task 10 with X gate applied after Z gate - the tasks in this kata ask for the exact state change rather than "up to a global phase" so that solution doesn't pass the test.

TonyHoldroyd:
Ah,thanks for letting me know Mariia. I'll edit and retry. That makes sense

PR #724, opened 2022-01-11
Add Workbook for tutorials/RandomNumberGeneration Kata

ssc1729:
Hi there, 

Add workbook for tutorials/RandomNumberGeneration kata. Addressing one of subtask in https://github.com/microsoft/QuantumKatas/issues/542. Welcoming all feedback!

Thanks!


tcNickolas (approved):
Looks great! I did a bit of polish, the biggest one in exercise 3 - your solution covered the more general case that you can use in exercise 5, and in this one I want to show a simpler solution, so that the learner gets the gradual progress.

Thank you!

PR #688, opened 2021-11-16
Switched to a new dump functionality exposed by QuantumSimulator

kuzminrobin:
This PR is a stage two of the three below (fixes [this failure](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=35503&view=logs&j=7c911cd3-259e-5f9c-1871-e3ecdf8c19e2&t=49e51bed-911d-557f-68be-a68a81a3af9b&l=972) in `iqsharp` [PR](https://github.com/microsoft/iqsharp/pull/544)):
* Preparing the QuantumSimulator for unifying the Dump(). Providing the new dump functionality on the `qsharp-runtime` repo side ([PR](https://github.com/microsoft/qsharp-runtime/pull/867)).
* Redirecting the `IQSharp` ([PR](https://github.com/microsoft/iqsharp/pull/544)) and `QuantumKatas` repo to the new dump functionality in `qsharp-runtime` repo.
* Cleaning-up the commented out code on the `qsharp-runtime` repo side.

PR #727, opened 2022-01-12
Switched to a new dump functionality exposed by IQ#.

kuzminrobin:
Simplified the Dump functionality for QuantumSimulator and the functionality use in IQSharp.
This is a part of the set of 3 PRs:
[Q#RT side](https://github.com/microsoft/qsharp-runtime/pull/898),
[IQ#](https://github.com/microsoft/iqsharp/pull/562),
[Katas](https://github.com/microsoft/QuantumKatas/pull/727).


cgranade (approved):
This looks good to me, but would be good to get @tcNickolas' review as code owner as well. One thing I'll point out quickly, though, is that as written in this PR, the `%kata` magic command won't support histogram views for displayable states. I think that was already the case before this PR, so not a regression so much as a known limitation.

tcNickolas:
@kuzminrobin I'm not very familiar with the katas magic part of code, @anjbur usually helps me with reviews on them. Can you please provide a bit of context on this change? Are there going to be any user-facing behavior changes? What about @cgranade's  comment on #688?

(Also, it looks like this PR supersedes #688 - could you please close the other PR to keep things more organized?)

kuzminrobin:
> Can you please provide a bit of context on this change? 

This change stops calling the functionality (in IQ#) that is being removed as a simplification of the Dump functionality in QuantumSimulator. 
> Are there going to be any user-facing behavior changes?

No. Nothing that the user's code could be relying on.
> What about @cgranade's comment on #688?

I added a reply there that the note has been resolved in this repeated attempt.
> could you please close the other PR

Done. Thanks for the reminder! ;-)

tcNickolas (approved):
@cgranade Right, [this bug](https://github.com/microsoft/iqsharp/issues/450)! I thought initially that your comment was about the measurement probability bars shown in regular DumpMachine output, thank you for clarifying!

That was my one concern, I'll trust you folks on the syntax of the change :-) Approving, I'll let you update the branch to latest and merge when the time is right (not sure if it needs to go in sync with the other two PRs)

utilities/Microsoft.Quantum.Katas/KataMagic.cs
```diff
@@ -206,8 +207,8 @@ public virtual bool Simulate(OperationInfo test, OperationInfo userAnswer, IChan
         /// Creates the instance of the simulator to use to run the Test 
         /// (for now always CounterSimulator from the same package).
         /// </summary>
-        public virtual SimulatorBase CreateSimulator(IChannel channel) =>
-            new CounterSimulator().WithJupyterDisplay(channel, ConfigurationSource);
+        public virtual SimulatorBase CreateSimulator() =>
+            new CounterSimulator();
```
DmitryVasilevsky, line 211: Cleans it up nicely!

DmitryVasilevsky (approved):

PR #739, opened 2022-01-26
Updates to new array creation syntax, part 2

tcNickolas:
This change covers cases not handled by the Q# Formatter.


ScottCarda-MS (approved):
Nice! I'm really liking this new array syntax. Using the `+=` is especially clean looking.

PR #744, opened 2022-02-03
Removing paths from qdk-sync trigger

anpaz:
Currently we only trigger a qdk build when changes on utilities, but sometimes we've seen problems when changes to Katas.
This removes the paths so every commit to main triggers a new qdk build.
Also adding the `workflow_dispatch` event so we can trigger this on demand with other branches.

### DIFF
--- a/.github/workflows/qdk-sync.yml
+++ b/.github/workflows/qdk-sync.yml
@@ -1,11 +1,10 @@
 name: Sync QDK repos
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
-    paths:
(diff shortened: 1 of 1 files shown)


anpaz:
Please merg this to main at your earliest convenience.

tcNickolas:
To double-check, this change doesn't affect the CI builds, right? Only what happens after the PR is merged to main

anpaz:
Yeah correct. Only what happens after merging to main.

tcNickolas (approved):
Sounds good!

PR #747, opened 2022-02-08
Enable Azure Quantum target selection for online job submissions

### LINKED ISSUE #741: Enable Azure Quantum job submissions from Binder
`%azure.target` cell in [Part IV of the Deutsch-Jozsa tutorial](https://github.com/microsoft/QuantumKatas/pull/734) fails to execute on Binder. `%azure.connect` cell allows the user to connect to Azure and shows the list of targets, but attempting to set a specific target fails  with the `Unable to find package 'Microsoft.Quantum.Providers.IonQ'.` error.

On Binder, we restore NuGet packages when building the Docker image, and then prohibit to restore them when the VM is running (to speed up IQ# start time). We need to make sure Providers.IonQ package is loaded at image build time and is available at runtime.

* [Adding this package to the project file](https://github.com/microsoft/QuantumKatas/pull/735) fixed the problem (with this change the notebook runs successfully), but breaks the E2E builds.
* I couldn't [download the package using `nuget install`](https://github.com/microsoft/QuantumKatas/compare/mariia/install-nuget-packages?expand=1) because Binder doesn't recognize `nuget` as a command available to it.

IsraelMiles:
Hosted notebooks don't support Katas, these changes add providers as part of a standalone project file in order to allow target providers to be assigned in an online environment - fixes #741


azure-pipelines.yml
```diff
@@ -6,6 +6,7 @@ pr:
 - main
 - feature/*
 - release/*
+- imiles/*
```
tcNickolas, line 9: This change is just for testing the e2e builds, not for the final update that will go to the main branch, right?

tcNickolas (approved):
I've verified that this change allows part IV of ExploringDeutschJozsaAlgorithm tutorial to submit jobs to IonQ, so it looks good from that perspective. Once you've verified that this change doesn't break the main e2e builds, I think we can merge this change. (And Wednesday morning is a good time, so that we have time to revert if anything does go wrong!)

Thank you for working on this!

azure-pipelines.yml
IsraelMiles: That's right, it's removed now. 🙂

PR #756, opened 2022-02-25
Basic Gates Workbook Task 2.4

tonyholdroyd112:
Second attempt


tcNickolas (approved):
Looks good! I did a bit of cleanup, mostly fixed the order of \theta and \eta in the column vector representation of the state - an easy typo to make :-)

Thank you!

PR #764, opened 2022-03-16
Adopt .NET6 + VS2022 in Quantum Development Kit

ricardo-espinoza:
With this change, we're migrating the main release train of the QDK to the most recent Long Time Support version of the .NET framework. For details about this change, refer to the original issue [Q#C:1224](https://github.com/microsoft/qsharp-compiler/issues/1224).

As part of this change, we're:

- Re-targeting all .NetCoreApp3.1 binaries to .NET6.0
- Updating Docker images, samples and templates.
- Libraries using .NetStandard2.1 are not affected by this change.
- The minimum supported .NET version in the QDK will also be updated from 3.1 to 6.0
- The Visual Studio extension will now target Visual Studio 2022.

**Important: This is a multi-repo change, and all changes will need to be merged concurrently.**


tcNickolas:
@ricardo-espinoza <https://github.com/ricardo-espinoza> After this PR the
main branch of the Katas requires .NET6 and the tooling, while all the
installation instructions in our documentation mention only .NET 3.1.
Please prepare a follow-up PR to the Katas README to include the
information about the update and QDK setup instructions for the new
version, including the minimal supported versions of the VS Code and Visual
Studio, installation of the extensions for .NET6, and any other
installation steps that are different compared to the ones listed in our
official documentation.

Additionally, please revert the change to scripts/Update-QDKVersion.ps1; we
do not hardcode the QDK version in the script, but rather provide it as a
command-line parameter.

PR #765, opened 2022-03-18
Adapt to new async handling of compilation calls

cgranade:
This PR adapts the `%kata` magic command to include changes to asynchronous handling of compilation calls (https://github.com/microsoft/iqsharp/pull/615), and fixes a small typo in `Update-QDKVersion.ps1`.


tcNickolas (approved):
Looks good, I think you can keep one more project on the existing QDK version rather than update it.

Which beta version is this - is it the one built from https://github.com/microsoft/iqsharp/pull/615?

tcNickolas:
Hmm, I'm not sure where did my in-code comment go, the project that can be kept on existing beta rather than the new one is `utilities/DumpUnitary/DumpUnitary.csproj` - it's a separate tool that shouldn't be affected.

cgranade:
> Looks good, I think you can keep one more project on the existing QDK version rather than update it.

Sounds good, I'll revert that file as well, then.

> Which beta version is this - is it the one built from [microsoft/iqsharp#615](https://github.com/microsoft/iqsharp/pull/615)?

That's correct. The corresponding e2e build included an earlier version of this PR (without the version update) to check that everything builds correctly together.

PR #766, opened 2022-03-20
Fix quickref sign error

japanavi:
Fixed the sign error in Math Reference -> Complex Arithmetic -> Division. 

I was unable to compile this because I did not have access to the fonts used. As discussed via email, please compile and add the PDF this renders, @vivanwin. Thank you!


tcNickolas (approved):
Looks good - thank you for your vigilance, @japanavi!

PR #777, opened 2022-04-06
Remove deprecated `using` binding scope syntax

### LINKED ISSUE #776: Q# compiler reports that `using y` grammar is being deprecated
In the Grover's Search tutorial the following solution may not work permanently -- there may be other instances of this class of issue replicated elsewhere. 

It's worth noting too that syntax has been shuffled around on this twice addressing both :
- modifying the keyword: `using` -> `use` 
- elimination of the parenthesis surrounding the assignment operator `(eval = Qubit())` -> `eval = Qubit()`

https://github.com/microsoft/QuantumKatas/blob/14071e2a044879402bacd8d3a39f7a26ae686603/tutorials/ExploringGroversAlgorithm/ExploringGroversAlgorithmTutorial.ipynb?short_path=c4fa011#L344

meltyness:
https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/statements/quantummemorymanagement#use-statement reflects the correct syntax.

Resolves #776


tutorials/Qubit/Qubit.ipynb
```diff
@@ -163,7 +163,7 @@
     "> Before Q# 0.15 the syntax for qubit allocation was different:\n",
```
tcNickolas, line 163: Hmm, looking at this code block closer, it is actually a historical note on how the syntax looked like before the new `use` syntax, so it does make sense to use `using` here... Q# 0.15 was released exactly a year ago, so it probably makes sense to remove the historical part of the note, and switch it to describe the alternative current syntax
```suggestion
    "> An alternative syntax for qubit allocation defines the scope in which the allocated qubits are available explicitly:\n",
```

tcNickolas: And then we'll need to update line 170 as well, `'using' block` -> `'use' block` (GitHub doesn't let me leave a comment there directly)

tcNickolas (changes requested):
Thank you for reporting and fixing this! Looks good, just one small suggestion to switch the occurrence in the Qubit tutorial from a historical note to an alternative syntax description.

tcNickolas (approved):
Looks good! I did the last change, replaced the last reference to `using` at line 170.

Thank you!

PR #789, opened 2022-05-25
Update global.json file

IsraelMiles:
We need to update the qdk version value in global.json in addition to other files. This PR does so while also refactoring duplicate logic into a single function.

### DIFF
--- a/scripts/Update-QDKVersion.ps1
+++ b/scripts/Update-QDKVersion.ps1
@@ -41,48 +41,38 @@ $csFiles | ForEach-Object {
     } | Set-Content -Encoding UTF8 $_.Path
 }
 
+function findAndReplace() {
+    param (
+        [string]$stringToReplace,
+        [string]$targetPath
+    )
+
+    (Get-Content -Path $targetPath) | ForEach-Object {
+        $isQuantumPackage = $_ -match $jsonString
+        if ($isQuantumPackage) {
+            $_ -replace $Matches.oldVersion, $Version
+        } else {
+            $_
+        }
+    } | Set-Content -Path $targetPath
+}
+
 $dockerString = "FROM mcr.microsoft.com/quantum/iqsharp-base:$versionRegex"
 $dockerPath = Join-Path $katasRoot "Dockerfile"
-(Get-Content -Path $dockerPath) | ForEach-Object {
-         $isQuantumPackage = $_ -match $dockerString
-         if ($isQuantumPackage) {
-             $_ -replace $Matches.oldVersion, $Version
-         } else {
-             $_
-         }
-    } | Set-Content -Path $dockerPath
-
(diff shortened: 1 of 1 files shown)


scripts/Update-QDKVersion.ps1
```diff
@@ -41,48 +41,38 @@ $csFiles | ForEach-Object {
...
 }
 
+function findAndReplace() {
+    param (
+        [string]$stringToReplace,
+        [string]$targetPath
+    )
+
+    (Get-Content -Path $targetPath) | ForEach-Object {
+        $isQuantumPackage = $_ -match $jsonString
```
tcNickolas, line 51: I suspect this should use the parameter rather than the global variable. When you tested it, did this script replace all the occurrences of the old version in all files?
```suggestion
        $isQuantumPackage = $_ -match $stringToReplace
```

scripts/Update-QDKVersion.ps1
```diff
@@ -41,48 +41,38 @@ $csFiles | ForEach-Object {
     } | Set-Content -Encoding UTF8 $_.Path
 }
 
+function findAndReplace() {
+    param (
+        [string]$stringToReplace,
```
tcNickolas, line 46: I'd call this variable `$patternToSearch`, since we're not replacing the whole string, just parts of it

scripts/Update-QDKVersion.ps1
```diff
@@ -41,48 +41,38 @@ $csFiles | ForEach-Object {
     } | Set-Content -Encoding UTF8 $_.Path
 }
 
+function findAndReplace() {
```
tcNickolas, line 44: I'd make the name a bit more descriptive, since it's not _quite_ find-and-replace you see in IDEs normally
```suggestion
function FindPatternAndReplaceQDKVersion() {
```

tcNickolas (changes requested):
I made a pass and left a couple of comments. Thank you!

scripts/Update-QDKVersion.ps1
IsraelMiles: Thanks for the feedback! I've added your suggestions and started a new pipeline run 🙂

scripts/Update-QDKVersion.ps1
```diff
@@ -18,7 +18,7 @@ param(
         
     .EXAMPLE
     
-        PS> ./Update-QDKVersion.ps1 -Version 0.24.210930
+        PS> ./Update-QDKVersion.ps1 -Version test-123
```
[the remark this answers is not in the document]
IsraelMiles: Just fixed, thank you for being a detailed reviewer 🙂

tcNickolas (approved):
Looks good, thank you so much for doing this!

PR #793, opened 2022-06-17
Update iqsharp-base version

anjbur:
Update the base image version to pull in security patches.


tcNickolas (approved):
Looks good, and I double-checked that it builds on Binder. Thank you!

PR #794, opened 2022-06-19
Part 1 of RippleCaryAdder workbook

Parisha-Agrawal:
Solutions for tasks 1.1 - tasks 1.3 of Ripple Carry Adder kata as per discussion in https://github.com/microsoft/QuantumKatas/issues/542#issuecomment-1145002068

### DIFF
--- a/RippleCarryAdder/RippleCarryAdder.ipynb
+++ b/RippleCarryAdder/RippleCarryAdder.ipynb
@@ -82,6 +82,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


tcNickolas (approved):
Looks good! I did a bit of cleanup (a couple of missing dots in the ends of sentences and whitespace seem not worth another round of review), and added a note to clarify that the workbook reasons about classical states - otherwise the reader could be confused about the relationship between the quantum states in the tasks and the classical bits in the solutions.

Thank you! Looking forward to the next parts! I'll be travelling next week so if you send a PR at that time, I'll get to it in July.

Parisha-Agrawal:
Thanks @tcNickolas  for quick turnaround on the PR. Looking forward to contribute the next parts. I am planning to send the next part in July for your review :-)

PR #801, opened 2022-07-11
[README] Add instructions about updating IQ# kernel

Manvi-Agrawal:


tcNickolas (approved):
Good point! I also took the liberty to update a couple of other install links, since I noticed that they were pointing to general pages and not instructions for the specific environment.
Thank you!

Issue #799, opened 2022-07-09
Sample solutions for task 1.14 of the Superposition katas.


ivan-shaporov:
https://github.com/microsoft/QuantumKatas/blob/cd7b60ea618512b0012f16f631c6c780d623841a/Superposition/Workbook_Superposition_Part2.ipynb?short_path=8af21c1#L622

I am looking at the sample solutions for task 1.14 of the Superposition katas. How about another option:
```q#
operation AllStatesWithParitySuperposition (qs : Qubit[], parity : Int) : Unit {
    for i in 1 .. Length(qs) - 1 {
        H(qs[i]);
        CNOT(qs[i], qs[0]);
    }

    if (parity == 1) {
        X(qs[0]);
    }
}
```

It is simpler than what is suggested and passes the tests. Isn't it correct?

tcNickolas:
It looks correct, yes! Passing the tests is a good sign, and I think you can also demonstrate its correctness recursively, since the for loop adds one qubit at a time and keeps the parity of the states even. 

Do you want to add it to the workbook as an alternative solution? It's harder for me to come up with solutions which use gates with a "later" qubit as control and an "earlier" one as target, so this solution can be interesting for learners who have the same "forward-CNOTs" bias :-)

ivan-shaporov:
Thank you @tcNickolas for looking into this. I've created a PR #802

PR #802, opened 2022-07-11
Add alternative sample solution for the Superposition kata task 1.14.

ivan-shaporov:
For issue #799.
Add alternative sample solution for the Superposition kata task 1.14.


tcNickolas (changes requested):
I think the solution description needs more details. It took me a bit of thinking to figure out how it works from just looking at it yesterday. The learner going to the workbook means that they struggled with this problem, so a more detailed step-by-step solution will be very helpful to them.

I'd describe the effect of the for loop using the recursive logic - look at the base case to show what happens there, and then what happens when you add one more qubit, that the loop always prepares superposition of all basis states with parity 0. Then the last if just puts the finishing touches :-) 

Thank you!

tcNickolas (approved):
Looks good! I took the liberty to add a bit more detail on how the for loop works, I think it can be helpful for the learner.

Thank you!

ivan-shaporov:
Alright, thank you.

Issue #803, opened 2022-07-18
Mistake in multi-qubit measurement tutorial exercise 2 calculations.


ivan-shaporov:
https://github.com/microsoft/QuantumKatas/blob/cf3d7e3cbae6f5a0bfd0cf67381ce613d7412383/tutorials/MultiQubitSystemMeasurements/Workbook_MultiQubitSystemMeasurements.ipynb?short_path=78ffc93#L97-L101

I repeated the calculations and got a different result with the amplitudes of the last two terms swapped:

$\\newcommand{\\ket}[1]{\\left|#1\\right>}$,
$\\newcommand{\\bra}[1]{\\left<#1\\right|}$
$$\\ket \\psi =  \\frac{2}{3}\\ket {00} + \\frac{1}{3} \\ket {01} + \\frac{2}{3}\\ket {11} = $$ 
$$= \\frac{2}{3} \\big[ \\frac{1}{\\sqrt{2}}\\big(\\ket{+} + \\ket{-}\\big) \\otimes \\frac{1}{\\sqrt{2}} \\big(\\ket{+} + \\ket{-}\\big) \\big] + $$ 
$$+ \\frac{1}{3} \\big[ \\frac{1}{\\sqrt{2}}\\big(\\ket{+} + \\ket{-}\\big) \\otimes \\frac{1}{\\sqrt{2}} \\big(\\ket{+} - \\ket{-}\\big) \\big] + $$ 
$$+ \\frac{2}{3} \\big[ \\frac{1}{\\sqrt{2}}\\big(\\ket{+} - \\ket{-}\\big) \\otimes \\frac{1}{\\sqrt{2}} \\big(\\ket{+} - \\ket{-}\\big) \\big] = $$ 
$$= \\frac{1}{3} \\big[ \\big(\\ket{+} + \\ket{-}\\big) \\otimes \\big(\\ket{+} + \\ket{-}\\big) \\big] + $$ 
$$+ \\frac{1}{6} \\big[ \\big(\\ket{+} + \\ket{-}\\big) \\otimes \\big(\\ket{+} - \\ket{-}\\big) \\big] + $$ 
$$+ \\frac{1}{3} \\big[ \\big(\\ket{+} - \\ket{-}\\big) \\otimes \\big(\\ket{+} - \\ket{-}\\big) \\big] = $$ 
[...]

PR #804, opened 2022-07-18
Fix calculations for Multi-qubit measurement tutorial exercise 2.

ivan-shaporov:
#803 

Doing exercise 2 in Multi-qubit measurement tutorial I am getting a result that is different from what is in the answer.

The final result table also seems wrong having ++ term twice.


tcNickolas (approved):
Yes, it does look like there was a mistake, the amplitudes of +- and -+ should have the same absolute value, since 01 is the only basis state that will end up contributing to them. Nice catch, thank you for your vigilance! 

We could also automate this computation using some Q# code; I opened #807 to describe the idea.

PR #805, opened 2022-07-19
Enable multicell solution tests in CHSHGame Workbook

Manvi-Agrawal:
### Approach
- We can enable multicell tests in the workbook by using `raises-exception` cell tag for tests that are guaranteed to fail for first run. Because this cell tag ignores any exceptions encountered during the cell execution.
- However, we can only enable the Workbook tests. Kindly refer to #806 for more details.

### Ref links
#539 : for more details about exclusion of multicell solution tests


Manvi-Agrawal:
@tcNickolas thoughts?

tcNickolas (approved):
Agreed, this makes sense. (And apologies for the delay reviewing, I needed to recall what we were doing with the notebooks validation, and that took a while :-))

Thank you!

Manvi-Agrawal:
No worries @tcNickolas. I understand that you have limited bandwidth to work on Katas. Thanks for taking a look at this.

PR #808, opened 2022-07-26
Update Katas to 0.25.2207 release

ricardo-espinoza:


tcNickolas:
The set of files updated looks good. I assume you'll be updating it to the release version instead of the beta next?

ricardo-espinoza:
> The set of files updated looks good. I assume you'll be updating it to the release version instead of the beta next?

Thanks. Yes, now the version has been updated to the released one.

tcNickolas (approved):
Looks good, and I've checked that Binder runs fine after the change. Thank you!

PR #810, opened 2022-07-28
[Superposition] Fix typo in URL

remilvus:
At the end of the [Superposition workbook](https://github.com/microsoft/QuantumKatas/blob/main/Superposition/Workbook_Superposition.ipynb) there is a typo in the circuit image URL.


tcNickolas (approved):
Ah, good catch! My work OS is Windows which is not sensitive to filename case in this scenario, so this kind of issues are harder to notice. Thank you!

Issue #807, opened 2022-07-23
[MultiQubitSystemMeasurements] Add code to Exercise 2 solution to show state in Pauli X basis


tcNickolas:
Follow up to #804: it is possible to verify the math computation for the state in the Pauli X basis by preparing the given state and then applying an H gate to each qubit - the amplitudes of the computational basis states in the resulting state will match the amplitudes of the Pauli X basis states in the original state. I was tempted to write this code when reviewing the pull request, so it could be a good thing to add to the workbook solution.

remilvus:
I can work on this task

tcNickolas:
Sounds great! It should be rather straightforward, but let me know if you have any questions!

PR #812, opened 2022-07-29
Add Q# based solution

remilvus:
This PR addresses #807  
It includes an additional solution and explanation for Exercise 2. The new solution is verifying the existing analytical solution with Q# code.


tcNickolas (approved):
This looks great - probably more detailed than I would've done :-)
I added one sentence to clarify what the transformation between the basis states is (just in case), and a quick DumpMachine before the transformation to show the amplitudes of the state |𝜓⟩ before it changes, to double check that we got the right state.

Thank you!

PR #813, opened 2022-07-29
Fix typo in Superposition workbook part II

remilvus:
I've changed "type casing" into "type casting"


tcNickolas (approved):
Nice catch! Thank you!

Issue #809, opened 2022-07-28
[Qubits and Gates] Fix math mode rendering in browser


remilvus:
Hi, I've noticed that equations written in math mode are not always rendered correctly in a browser. Fixing this would make the learning experience better for people that use separate devices for reading and solving katas.

### Example of the problem
Equations written in consecutive lines are rendered as one-line equation. 

In the [SingleQubitGate kata](https://github.com/microsoft/QuantumKatas/blob/main/tutorials/SingleQubitGates/SingleQubitGates.ipynb) under the "Matrix Representation" section the equations for $X|0\rangle$ and $X|1\rangle$ are displayed on one line instead of two.
The problem is also visible in [Complex Arithmetic kata](https://github.com/microsoft/QuantumKatas/blob/main/tutorials/ComplexArithmetic/ComplexArithmetic.ipynb) under the "Imaginary numbers" section. There three equations are displayed on one line.

I can fix this by dividing the equations into multiple math sections instead of using `\\` to generate a new line. For example `$$eq1 \\ eq2$$` would be transformed into `$$eq1$$ $$eq2$$`. [Here you can view example of my fix](https://github.com/remilvus/QuantumKatas/blob/main/tutorials/SingleQubitGates/SingleQubitGates.ipynb).

### Technical details
I've verified that this rendering problem exists on Edge, Firefox and Chrome.

Changing the colour of an equation is also broken in the browser (See: SingleQubitGates -> Ket-bra decomposition).

-------------------

If you don't have any objections to this fix, I can search and fix the rendering problems also in other notebooks. If I do that it would be nice to know which katas are equation-heavy, so if you know please share.

tcNickolas:
This can be a good thing to fix indeed. We've always assumed that the main use case for the katas is interactive, since the project started with the katas, and added tutorials and workbooks that had valuable reading material only much later! So we never spent time on renderings in GitHub preview mode. 

Some things might just not be supported - I'm not sure GitHub previewer handles color at all? And there might be more broken things, for example, Ket-bra decomposition from your example shows up as several notes broken up by > signs rather than a single unbroken note.

If you have time to track down this kind of issues and figure out how to fix them without breaking the rendering in interactive mode, that will be great!

[...]

PR #814, opened 2022-07-29
 Fix math mode rendering in browser

remilvus:
This addresses #809


tcNickolas (approved):
Looks great! I fixed two pre-existing typos in MultiQubitGates, and re-saved Workbook_BasicGates to remove the cell IDs generated automatically to make the scope of changes cleaner for that file.

Thank you for doing this!

PR #819, opened 2022-08-06
Cleanup solution for task 1.7 in QFT Workbok

Manvi-Agrawal:
We have two solution blocks for task 1.7 with second one containing extra contents when compared to the first one. Hence, removing the first solution block.


tcNickolas (changes requested):
This is weird... I suspect what happened here is that we tried to split the longer solution cell in two cells to make editing easier, but then forgot to remove the matching content from the longer cell. I think it makes sense to do the change that way, to end up with two shorter cells. Can you do this change, please?

Thank you for catching this!

Manvi-Agrawal:
@tcNickolas , thanks for a quick turn around on this. I didn't imagine that it could be happening because we wanted to split a longer solution into 2 cells. I have done the change requested by you.

tcNickolas (approved):
This just seems like the most likely explanation - could also be a weird copy-paste issue :-)

Thank you!

PR #820, opened 2022-08-06
The current ripple carry adder reference solution could be a bit more…

abrassel:
Instead of manually applying the adjoints to clean up afterwards, we can use Q#'s very nice `within {...} apply {...}` syntax to do it for us.


RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -73,28 +73,24 @@ namespace Quantum.Kata.RippleCarryAdder {
     // Task 1.7. N-bit adder
     operation ArbitraryAdder_Reference (a : Qubit[], b : Qubit[], sum : Qubit[], carry : Qubit) : Unit is Adj {
         let N = Length(a);
-        if (N == 1) {
-            LowestBitSum_Reference(a[0], b[0], sum[0]);
-            LowestBitCarry_Reference(a[0], b[0], carry);
+        if N == 1 {
+            OneBitAdder(a[0], b[0], sum[0], carry);
```
tcNickolas, line 77: You have to use `_Reference` versions of all tasks you use in the reference solutions, since the regular versions are empty in the project (you can see that the tests are failing on this kata in the CI builds)

RippleCarryAdder/ReferenceImplementation.qs
```diff
@@ -73,28 +73,24 @@ namespace Quantum.Kata.RippleCarryAdder {
...
         }
         else {
-            use internalCarries = Qubit[N-1];
-            LowestBitSum_Reference(a[0], b[0], sum[0]);
-            LowestBitCarry_Reference(a[0], b[0], internalCarries[0]);
-                
-            for i in 1 .. N-2 {
-                HighBitSum_Reference(a[i], b[i], internalCarries[i-1], sum[i]);
-                HighBitCarry_Reference(a[i], b[i], internalCarries[i-1], internalCarries[i]);
+            use carries = Qubit[N-1];
```
tcNickolas, line 80: I would keep the name `internalCarries`, to make it easier to distinguish it from the `carry` parameter. This will also help GitHub identify the differences between the old code and the new one easier

tcNickolas (changes requested):
Wow, I didn't realize that this solution predates the within...apply construct! Or nearly predates it :-)

I left a couple of comments - I think once you address them, the change should be good, and it should pass the CI build.

Thank you!

RippleCarryAdder/ReferenceImplementation.qs
abrassel: Thanks!

abrassel (commented):
I added the changes you requested!

abrassel:
It looks like the failures are for graph coloring?  Maybe spurious.

tcNickolas (approved):
Looks good! CI failures on a different kata shouldn't be related, I re-launched CI build with the branch update to the latest main.

Thank you!

PR #822, opened 2022-08-09
RippleCarryAdder Workbook : Task 2.1 and 2.2

Parisha-Agrawal:
@tcNickolas Here is the solution for task 2.1 and 2.2 in the Ripple Carry Adder Workbook. Please review.


tcNickolas (approved):
Looks great, thank you!

PR #826, opened 2022-08-16
[Superposition] Restore link to Bloch sphere presentation

tcNickolas:
We had to remove this link in https://github.com/microsoft/QuantumKatas/pull/751 and replace it with a link to Wikipedia, but now I've found the new home of the original presentation. It's still the best presentation on Bloch sphere I've ever seen, so bringing it back.


vivanwin (approved):
This looks good thank you!

Issue #818, opened 2022-08-05
Late introduction of `ControlledOnInt` operation


remilvus:
### Problem
If someone is flawlessly solving notebooks in order that the "Learning Path" suggests, then they will encounter the `ControlledOnInt` operation for the first time in [Oracles tutorial](https://github.com/microsoft/QuantumKatas/blob/main/tutorials/Oracles/Oracles.ipynb). 
This is a quite useful operation, that is used for example in the Superposition workbook (solution to task 1.12). 

### Proposed solution
If workbooks are supposed to be skipped when someone can solve a task, then I think that this operation can appear earlier. Maybe it can be introduced as a hint in task 1.12, or even as a separate task in which `ControlledOnInt` needs to be implemented?

tcNickolas:
That's a good point, indeed Superposition task 1.12 uses it in the solution (as well as ControlledOnBitString) but doesn't offer a hint for it.

Do you want to open a pull request adding a hint to Superposition task 1.12 mentioning ControlledOnInt and ControlledOnBitString as useful library functions to use? A separate task wouldn't really work in the context of Superposition kata, since all the tasks there say "prepare this state" rather than "implement this unitary transformation".

remilvus:
Yes, I would like to do that.

When mentioning adding this as a separate task, I was thinking it could be in one of the katas that come before Superposition. For example, Basic Gates Kata looks like a good candidate, as controlled gates are introduced there. 
In my opinion, it's good to have both a hint and a task, especially if `ControlledOnBitString` operation is included. What do you think? Dividing new concepts into separate tasks seems like a good strategy, but at the same time it generates an infinite number of tasks in the limit.

tcNickolas:
Exercise 5 of the MultiQubitGates tutorial covers `ControlledOnBitString` function in quite a lot of detail, so we probably don't need an infinite number of tasks :-) BasicGates can be a good place for a task "Two-qubit gate - 4" (between the current tasks 2.3 and 2.4) introducing a zero-controlled NOT gate, as a simpler version of that exercise, and you can introduce both ControlledOn* functions in the hint/workbook for it. Maybe also point to that exercise for a more detailed version of the discussion.

PR #825, opened 2022-08-12
Fix auto cell id generation by downgrading nbformat to 4.4

Manvi-Agrawal:
Fixes #823

PR #824, opened 2022-08-10
[Superposition] Add a hint in Task 1.12

remilvus:
This PR addresses issue #818

PR #827, opened 2022-08-17
Add a task to BasicGates

remilvus:
This PR addresses #818


Manvi-Agrawal:
Hmm, I know this is still a work in progress but I am quite intrigued by cell IDs generated by jupyter on your system. This was fixed by me in https://github.com/microsoft/QuantumKatas/pull/825. 

I can see that your branch does not contain the fix. Could you please do the following?
- Merge branch main to your branch.
- Reopen Basic Gates Workbook and save it and see if IDs go away.

Also, could you please share nbformat version which you are using locally? Thanks

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "operation TwoQubitGate4 (qs : Qubit[]) : Unit is Adj+Ctl {\n",
+    "    // ... \n",
+    "}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "*Can't come up with a solution? See the explained solution in the [Basic Gates Workbook](./Workbook_BasicGates.ipynb#Task-2.3.-Two-qubit-gate---4).*"
```
Manvi-Agrawal, line 587: Possibly a typo?
```suggestion
    "*Can't come up with a solution? See the explained solution in the [Basic Gates Workbook](./Workbook_BasicGates.ipynb#Task-2.4.-Two-qubit-gate---4).*"
```

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "### Task 2.4. Two-qubit gate - 4\n",
+    "\n",
+    "**Input:** Two unentangled qubits (stored in an array of length 2) in an arbitrary two-qubit state ${\\color{blue}\\alpha} |00\\rangle + {\\color{blue}\\beta} |01\\rangle + \\gamma |10\\rangle + \\delta |11\\rangle$.\n",
+    "\n",
+    "\n",
+    "**Goal:**  Change the two-qubit state to ${\\color{red}\\beta} |00\\rangle + {\\color{red}\\alpha} |01\\rangle + \\gamma |10\\rangle + \\delta |11\\rangle$.\n",
+    "\n",
+    "> This task can be solved using one library function; as an exercise, try to express the solution using several (possibly controlled) Pauli gates.  \n",
+    ">\n",
+    "> If you can't come up with a solution, or if you have solved the task, read about the relevant library function: \n",
```
Manvi-Agrawal, line 558: Instead of this line, it can be reworded to a hint on lines of something like :
Interested in knowing about library function? Click here

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "  <details> \n",
+    "    <summary><b>Click here.</b></summary>\n",
+    "    `ControlledOnInt` function can generate a controlled gate, which is applied only when the control qubit is in the $|0\\rangle$ state. </br>\n",
+    "    For example, `ControlledOnInt(0, H)([qs[0]], qs[1])` applies the Hadamard gate to the `qs[1]` qubit, when the control qubit `qs[0]` is in the $|0\\rangle$ state.\n",
```
Manvi-Agrawal, line 562: I liked your explaination for `ControlledOnInt` function in https://github.com/microsoft/QuantumKatas/pull/824. 

This might create a misunderstanding because we are saying that  `ControlledOnInt` function is only applied when register is in $| 0 \rangle$ state. This is  not always true because I can pass one to `ControlledOnInt` function, and then this gate will be applied if control register is in state $|1\rangle$

```suggestion
    "    `ControlledOnInt` function can generate a controlled gate, which is applied when the control register is in a state corresponding to a non-negative integer</br>\n",
    "    Example 1: For integer 0,  `ControlledOnInt(0, H)([qs[0]], qs[1])` applies the Hadamard gate to the `qs[1]` qubit, when the control qubit `qs[0]` is in the $|0\\rangle$ state.<br>\n",
    "    Example 2: For integer 0,  `ControlledOnInt(0, H)([qs[0]], qs[1])` applies the Hadamard gate to the `qs[1]` qubit, when the control qubit `qs[0]` is in the $|0\\rangle$ state.<br>\n",
```

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "    </br></br>\n",
```
Manvi-Agrawal, line 563: Minor suggestion. <br> is an empty tag in html. [Doc link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br)
```suggestion
    "    <br><br>\n",
```

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "    Notice that the `ControlledOnInt` funciton requires the control qubit(s) to be passed as a register - this function generates a gate controlled by multiple qubits. The gate is applied if the register is in a state equivalent to a non-negative integer (0 in the example above). A similar function is `ControlledOnBitString` which will be introudced in [Multi-Qubit Gates tutorial](../tutorials/MultiQubitGates/MultiQubitGates.ipynb).\n",
+    "    </br></br>\n",
```
Manvi-Agrawal, line 565: Similarly here.

```suggestion
    "    <br><br>\n",
```

BasicGates/BasicGates.ipynb
```diff
@@ -546,7 +546,52 @@
...
+    "\n",
+    "> This task can be solved using one library function; as an exercise, try to express the solution using several (possibly controlled) Pauli gates.  \n",
+    ">\n",
+    "> If you can't come up with a solution, or if you have solved the task, read about the relevant library function: \n",
+    "  <details> \n",
+    "    <summary><b>Click here.</b></summary>\n",
+    "    `ControlledOnInt` function can generate a controlled gate, which is applied only when the control qubit is in the $|0\\rangle$ state. </br>\n",
+    "    For example, `ControlledOnInt(0, H)([qs[0]], qs[1])` applies the Hadamard gate to the `qs[1]` qubit, when the control qubit `qs[0]` is in the $|0\\rangle$ state.\n",
+    "    </br></br>\n",
+    "    Notice that the `ControlledOnInt` funciton requires the control qubit(s) to be passed as a register - this function generates a gate controlled by multiple qubits. The gate is applied if the register is in a state equivalent to a non-negative integer (0 in the example above). A similar function is `ControlledOnBitString` which will be introudced in [Multi-Qubit Gates tutorial](../tutorials/MultiQubitGates/MultiQubitGates.ipynb).\n",
```
Manvi-Agrawal, line 564: Possibly some typos here and a bit of rewording
```suggestion
    "    Notice that the `ControlledOnInt` function requires the control qubit(s) to be passed as a register. This function generates a gate controlled by multiple qubits. The gate is applied if the control register state corresponds to the specified bit mask (0 in the first example above). A similar function is `ControlledOnBitString` which will be introduced in [Multi-Qubit Gates tutorial](../tutorials/MultiQubitGates/MultiQubitGates.ipynb).\n",
```

BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1521,16 +1633,176 @@
...
+    "   \\color{red}\\beta\\\\\n",
+    "   \\color{red}\\alpha\\\\\n",
+    "   \\gamma\\\\\n",
+    "   \\delta\\\\\n",
+    "\\end{bmatrix}\n",
+    "$$\n",
+    "\n",
+    "This can be represented by a matrix:\n",
+    "$$\n",
+    "T = \\begin{bmatrix}\n",
```
Manvi-Agrawal, line 1684: Maybe we can use $U$ to represent this transformation. Because we have a `T gate` available in Q#. So, it might make it a little difficult for somebody who is aware of that gate to follow along this task.

Although its not widely used like $X$, $H$ gates. Here is [link](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.t) to the documentation of T gate.

```suggestion
    "U = \\begin{bmatrix}\n",
```

BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1521,16 +1633,176 @@
...
+    "\\text{CNOT} = \n",
+    " \\begin{bmatrix}\n",
+    "   1 & 0 & 0 & 0\\\\\n",
+    "   0 & 1 & 0 & 0\\\\\n",
+    "   0 & 0 & 0 & 1\\\\\n",
+    "   0 & 0 & 1 & 0\\\\\n",
+    " \\end{bmatrix}\n",
+    "$$\n",
+    "\n",
+    "We need a way to transfrom the $\\text{CNOT}$ gate into the $T$ gate.   \n",
```
Manvi-Agrawal, line 1704: Suggestion : Change `T` to `U`

```suggestion
    "We need a way to transform the $\\text{CNOT}$ gate into the unitary transformation represented by $U$.   \n",
```

BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1521,16 +1633,176 @@
...
+    "We remember that the Pauli X gate flips the state in the single qubit case. Here we need to use a 2-qubit version of this gate, which would affect only the second qubit. We conclude, that the idenity gate needs to be used on the first qubit. In the end, the required gate is a tensor product: $(I \\otimes X)$.\n",
+    "\n",
+    "We validate that composition of $(I \\otimes X)$ and the $\\text{CNOT}$ gate produces the required $T$ gate:\n",
```
Manvi-Agrawal, line 1707: Suggestion : Change `T` to `U`.
```suggestion
    "We validate that composition of $(I \\otimes X)$ and the $\\text{CNOT}$ gate produces the required unitary transformation represented by $U$. \n",
```

BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1521,16 +1633,176 @@
...
+    "   0 & 0 & 0 & 1\\\\\n",
+    "   0 & 0 & 1 & 0\\\\\n",
+    "  \\end{bmatrix}\n",
+    "  =\n",
+    " \\begin{bmatrix}\n",
+    "   0 & 1 & 0 & 0\\\\\n",
+    "   1 & 0 & 0 & 0\\\\\n",
+    "   0 & 0 & 1 & 0\\\\\n",
+    "   0 & 0 & 0 & 1\\\\\n",
+    "  \\end{bmatrix} = T\n",
```
Manvi-Agrawal, line 1744: Suggestion : Change `T` to `U`.

```suggestion
    "  \\end{bmatrix} = U\n",
```

BasicGates/Workbook_BasicGates.ipynb
```diff
@@ -1521,16 +1633,176 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "80c6d622",
+   "metadata": {},
+   "source": [
+    "### Solution 2 - using a library function\n",
+    "\n",
+    "We observe that the task requires application of a Pauli X gate on the second qubit when the first qubit is in the $|0\\rangle$ state. This can be achieved with [`ControlledOnInt`](https://docs.microsoft.com/en-gb/qsharp/api/qsharp/microsoft.quantum.canon.controlledonint) library function.\n",
+    "\n",
+    "> Notice that the `ControlledOnInt` function creates a gate controlled by a register - not by a single qubit. A similar function (`ControlledOnBitString`) will be introudced in [Multi-Qubit Gates tutorial](../MultiQubitGates/MultiQubitGates.ipynb#Multi-Qubit-Gates)."
```
Manvi-Agrawal, line 1775: Possibly some typos here?

```suggestion
    "> Notice that the `ControlledOnInt` function creates a gate controlled by a register - not by a single qubit. A similar function (`ControlledOnBitString`) will be introduced in [Multi-Qubit Gates tutorial](../tutorials/MultiQubitGates/MultiQubitGates.ipynb#Multi-Qubit-Gates)."
```

Manvi-Agrawal:
Overall, your changes look good. I left some suggestions for you which you can apply by clicking on "Apply change" button to let Github do the changes automatically for you. Some of these changes are minor suggestion like fixing typos or some rewording. Thanks for working on this.

remilvus:
> Hmm, I know this is still a work in progress but I am quite intrigued by cell IDs generated by jupyter on your system. This was fixed by me in #825.
> 
> I can see that your branch does not contain the fix. Could you please do the following?
> 
> * Merge branch main to your branch.
> * Reopen Basic Gates Workbook and save it and see if IDs go away.
> 
> Also, could you please share nbformat version which you are using locally? Thanks

I'm using 5.3.0 version of nbformat locally, but now I see you downgraded it to the 4.x version. Thanks for the steps to fix this problem.

BasicGates/Workbook_BasicGates.ipynb
remilvus: Thanks for catching this! I was planning to introduce the gate as the **T**arget gate, but ended up not including this explanation.

Manvi-Agrawal:
Hi @remilvus, I was also on nbformat 5.3. On my system downgrading nbformat version fixed this problem. Then I opened the notebook in jupyter and  saved it again to fix this problem.

Can you please open the workbook in  jupyter locally, save it and see if it's still generating ids? If it does, can you please try running "Restart kernel and clear all outputs" and see if it fixes the problem.

BasicGates/BasicGates.ipynb
remilvus: Indeed

BasicGates/BasicGates.ipynb
remilvus: This is exactly why I've written "**can** generate" instead of "generates", but I see now that it generates confusion. Your version is better, but there are still some improvements needed (as both examples are the same).

Manvi-Agrawal: Thanks for catching this

Manvi-Agrawal: Could you just wait a minute? I am editing my suggestion so that you can apply it.Thanks.

BasicGates/BasicGates.ipynb
remilvus: About the rewording and typos: I mostly agree, although I think we should be consistent with using "a non-negative integer" instead of "the specified bit mask"

BasicGates/BasicGates.ipynb
remilvus: Ahh, I didn't expect such a quick reaction. I already applied the previous version :/

But I also wanted to reword one of your other suggestions, so I'm doing changes anyway

Manvi-Agrawal: That was in a different file

Manvi-Agrawal: I have edited my suggestion so that you can apply it.

BasicGates/BasicGates.ipynb
Manvi-Agrawal: I like "the specified bit mask" instead of "non-negative integer" as it specifies things more clearly. Only place I can see this in "SuperPosition" kata in hint added by you. So, you might want to rename it in this PR to be more consistent

remilvus:
> Hi @remilvus, I was also on nbformat 5.3. On my system downgrading nbformat version fixed this problem. Then I opened the notebook in jupyter and saved it again to fix this problem.
> 
> Can you please open the workbook in jupyter locally, save it and see if it's still generating ids? If it does, can you please try running "Restart kernel and clear all outputs" and see if it fixes the problem.

I didn't try clearing outputs but restarting the kernel didn't remove the IDs. Merging the main branch to downgrade nbformat helped by preventing the IDs from being generated after I manually removed them.

BasicGates/BasicGates.ipynb
remilvus: Thats a very good suggestion. I was to attached to the idea of teaching learners about this function 😅

Manvi-Agrawal:
> I didn't try clearing outputs but restarting the kernel didn't remove the IDs. Merging the main branch to downgrade nbformat helped by preventing the IDs from being generated after I manually removed them.

@remilvus, I am glad to know that my fix works. So, hopefully new contributors and the existing ones should not face this annoying issue of automatic cell ids if they have the changes in PR https://github.com/microsoft/QuantumKatas/pull/825.

 @tcNickolas, we got some validation here that it works. :-)

BasicGates/BasicGates.ipynb
remilvus: All of the relevant changes should be in [this commit] (https://github.com/microsoft/QuantumKatas/pull/827/commits/3473dabb461ef5a58b68c59f7662f9bfe2378a37)

BasicGates/BasicGates.ipynb
remilvus: I try to avoid using the "specified bit mask" to make a distinction between `ControlledOnInt` and `ControlledOnBitString`. I've added an additional sentence explaining that the "non-negative integer" specifies a bit mask. What do you think about such explanation?

remilvus: This way we are consistent with [documentation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.canon.controlledonint).
Now there is a "positive integer" specified there, but I've already created an issue to change that to a "non-negative integer".

remilvus: About navigating through the notebooks: 
When I was going through the "Quantum Computing Concepts: Qubits and Gates" section, I used the order from the "Learning path" as the order-in-which-notebooks-should-be-completed. It's probably a personal trait but I treat conclusion links as links-to-notebooks-that-you-may-want-to-have-a-look-at, even though it's stated that they are the next-learning-step.

It would be interesting to hear from @tcNickolas about this.

remilvus:
Thanks for all the suggestions @Manvi-Agrawal! Now the PR should be ready. 

I have some doubts about hiding the `ControlledOnInt` function under a hint, as teaching about this function is the main purpose of this task. 
On the other hand, a task explaining `ControlledOnInt` and asking the learner to use it would be a little too trivial to solve. In the current format, this task teaches mainly the concept behind a `ControlledOnInt(0, <gate>)` use case, which I think is more important than deeply explaining `ControlledOnInt`. The main reason is, that explaining exactly how the `ControlledOnInt` function works would almost be a repetition of the explanation for `ControlledOnBitString`.
It would be good to hear different opinions though 😃

Manvi-Agrawal:
@remilvus, I will let @tcNickolas answer the question about introducing the function as a hint.

I guess we have an item in our roadmap to have some Katas focused on introducing Q# to the learners.  There was an interest by "Tony Holdroyd" on this one in https://github.com/microsoft/QuantumKatas/issues/542#issuecomment-92416987.


Would you be willing to collaborate with him? For some initial guidance, you can read comments by "tcNickolas(Mariia)" after that question. @tcNickolas thoughts?

BasicGates/BasicGates.ipynb
Manvi-Agrawal: Regarding https://github.com/microsoft/QuantumKatas/pull/827#discussion_r949419262, sounds good.

remilvus:
@Manvi-Agrawal I'll be happy to help. I may not have enough time after the academic year start though.

Could you help me with figuring out why the checks are not passing? There are validation errors for the GraphColoring notebook, which I didn't modify. Have you seen similar errors before?

Manvi-Agrawal:
@remilvus, yes I vhave seen such issues before. Sometimes we might see failures in other Katas which we don't modify. This is because some of the tests in this repository have a passing probability of approx 99%( Eg: some tests in Graph coloring kata) So, once in a while they might fail. You can trigger CI by pushing an empty commit and it should pass second time

remilvus:
@Manvi-Agrawal thanks for the explanation. 
Wouldn't it be useful to mention that this can happen in the [CONTRIBUTING](https://github.com/microsoft/QuantumKatas/blob/main/.github/CONTRIBUTING.md#validating-your-changes) document? For example, under "Validating your changes" -> "3. Continuous integration"

tcNickolas (approved):
Apologies for the delay reviewing - I had a very intense August that left me no time for open-source work :-( 

The new task looks good! I did a bit of formatting edits - HTML tag `<details>` doesn't allow Markdown inside it, so I figured out it makes sense to keep this information in plain text, not hidden away. After all, introducing this function is the main reason why we're adding this task! Making this an easy task is good, this kata is more of a practice one than real problem-solving one, some of the tasks are even named for the gates the learner is supposed to use in them.

I've also added an extra check in the test to make sure that the reference solution matches the `ControlledOnInt` version and fixed a pre-existing issue in task 2.3 in which the notebook version of the task claims that the input qubits are unentangled (was really not worth opening a separate PR).

Thank you!

BasicGates/BasicGates.ipynb
tcNickolas: Catching up with this thread... Even before this change, the second part of the BasicGates kata listed MultiQubitGates tutorial as a resource. I think in this case we have either a loop of references or too many prerequisites for the BasicGates, since the kata covers both single- and multi-qubit basic gates, and the tutorials for them are split in two. We can clarify the final reference in the MultiQubitGates tutorial, saying "You should start with _the second part of_ the BasicGates kata", but other than that, I think it's fine as is.

Do you want to send a PR to add that clarification to the MultiQubitGates tutorial conclusion?

tcNickolas:
For the transient CI build failures, once we finish https://github.com/microsoft/QuantumKatas/pull/620, we should be able to address the root cause of many of the failures by switching the time-consuming tests to the new SparseSimulator. They are the tests that deal with Grover's search and quantum oracles, and this type of programs benefits from sparse simulation greatly, so timeouts should become much less frequent.

remilvus:
Thanks for making the last fixes!

PR #829, opened 2022-08-29
Fix display issue for `\bra` and `\ket` in VSCode

adrianleh:
Fixes an issue where the definition of `\bra` and `\ket` collides with loaded `braket` package in VSCode and causes error messages (that do not actually break functionality) to be displayed.

Further styles `bra` and `ket` more concisely with more accurate symbols


tcNickolas (approved):
I've checked that it works both locally in my browser and on Binder. I'll take your word that it improves the VS Code experience :-)

Thank you!

PR #835, opened 2022-09-16
Update URL for public doc repo

Bradben:


tcNickolas (approved):
Good catch, I completely forgot that the Katas reference the docs repo. Thank you!

PR #836, opened 2022-09-22
Updated LaTex rendering in solution of task 2.4

Amana-L:
Fixed links to task 2.3 and math mode rendering of the equations that followed. This follows issue #809.

### DIFF
--- a/Superposition/Workbook_Superposition_Part2.ipynb
+++ b/Superposition/Workbook_Superposition_Part2.ipynb
@@ -1087,7 +1087,7 @@
    "source": [
     "### Solution\n",
     "\n",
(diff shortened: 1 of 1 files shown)


Manvi-Agrawal:
Solution of Task 2.4 uses "ThreeStates_TwoQubits" function defined in solution of Task 2.3. So, the learner will need to run the solution of Task 2.3 in workbook to be able to run solution for task 2.4.

Maybe you can change link text from "task 2.3" to "solution to task 2.3" to avoid this confusion.


Feel free to do the changes as done by me in [this commit](https://github.com/Manvi-Agrawal/QuantumKatas/commit/8f47625d5798d5315ec1102317755054aa3d0b64).

Amana-L:
I see, thank you.

tcNickolas (approved):
This looks much better, indeed! Thank you for fixing this!

PR #838, opened 2022-10-17
Fix Typo in MultiQubitGate.ipynb

delight1019:
Hello.
I've fixed a formula of CNOT gate's eigen vector whick looks like a typo.


tcNickolas (approved):
Nice catch - wouldn't be much of an eigenvector if it changed under the gate application! Thank you!

PR #861, opened 2022-12-10
[Contributing Guide]Specify language for better rendering

Manvi-Agrawal:
### Change Description
text is formatted as per language syntax instead of plain black text

### Before
image

### After
image


tcNickolas (changes requested):
Good idea! Can you also add the xml language to the snippet with `<GeneratePackageOnBuild>` tag, so that all snippets in a specific language are annotated?

Thank you!

Manvi-Agrawal:
@tcNickolas , I have done the changes requested by you

PR #865, opened 2023-01-14
[README] Miscellaneous

Manvi-Agrawal:
Old Screenshot
Screenshot 2023-01-14 at 1 43 50 PM

New Screenshot
New rendering


tcNickolas (approved):
Good point, I guess the [!TIP] style propagated from documentation sites that render it in a different way from the GitHub rendering. Thank you!

Manvi-Agrawal:
Yes, @tcNickolas , you are right. I recently discovered that for the official Q# and C# documentation, we use DFM(Docfx Flavoured Markdown) which renders Tip quite nicely. This rendering has some extra features when compared to regular markdown

PR #875, opened 2023-02-22
Update the IQSharp.Jupyter Version

ScottCarda-MS:


samarsha (commented):
I think the katas don't like to use pre-release QDK versions?

tcNickolas:
That's correct, since the Katas are consumed by the user directly, alpha versions that can be trickier to obtain and might not be as stable as the released versions should not typically be used by the main branch. 

What is the goal of this update? Can it be done only in the `Microsoft.Quantum.Katas.csproj` file only, as the first commit tried to do, so that the parts of the project used by the end users remain on the released version?

tcNickolas:
We definitely cannot make the update like this - mcr.microsoft.com/quantum/iqsharp-base:0.27.47569-alpha is not published so Binder cannot build an image of this version for online use. I think making an alpha version usable by the external users would take a lot of steps we do in a release.

Is there a way in component governance to mark the issue as "will be fixed next week"? We're days away from the next QDK release, and if it includes this fix, we won't need to figure out a workaround for using the alpha version.

ScottCarda-MS:
Sure, I think we can hold off on updating the IQ# version referenced until after the release, then use the new release version of IQ# instead of the alpha.

tcNickolas (approved):
Sounds good, updating the M.Q.Katas project alone won't affect the user experience, and we'll pick up the release version next week. Thank you!

PR #877, opened 2023-03-02
[Ubuntu] Update instruction to run jupyter in new window

Manvi-Agrawal:
Hi @tcNickolas , recently I moved to Ubuntu machine and noticed that `start jupyter notebook <your notebook name>` didn't work. I found an equivalent command `gnome-terminal -- start jupyter notebook <your notebook name>` that serves the same purpose. Also attached the screenshots from new and the old wikis in PR description.

Updated new Readme
---
image
---

Old Readme
----
image
----

Updated Contributing guide
----
image
----

Old Contributing guide
----
image
----


tcNickolas (approved):
Good point! I knew `start` only works for Windows (and not universally too, some terminals don't know it), but somehow we never got to documenting this. Thank you!

PR #879, opened 2023-03-11
Removed broken link.

kuzminrobin:


tcNickolas (approved):
Weird that the blog went dead completely - I'd expect that it would be migrated to algassert.com... But even that page links to twistedoakstudios for this blog post, so I guess there's no newer version.

Thank you!

PR #896, opened 2023-10-10
Update README.md

mzanner:
updated README to add reference to new Azure Quantum katas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Introduction
 
-*** We are modernizing the quantum katas experience.   Please visit https://quantum.microsoft.com/experience/quantum-katas to try the new online Azure Quantum katas experience, with integrated assistance from Copilot in Azure Quantum. ***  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 # Introduction
 
+*** We are modernizing the quantum katas experience.   Please visit https://quantum.microsoft.com/experience/quantum-katas to try the new online Azure Quantum katas experience, with integrated assistance from Copilot in Azure Quantum. ***
```
tcNickolas, line 3: Three asterisks don't highlight the text. We want this to stand out, I'd recommend `>` in the beginning of the line (the note will look like the "Note that mybinder.org..." paragraph) or double asterisks (without spaces between them and text) to make it bold, or both

tcNickolas (changes requested):

mzanner (commented):
updated with recommended changes.

README.md
mzanner: updated per your recommendations!

tcNickolas (approved):

PR #900, opened 2024-04-14
Update Dockerfile

rohithreddykota:
fix iqsharp issues & add pkgs required for QuantumClassification tutorial

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN pip install -I --no-cache-dir \
         matplotlib \
         numpy \
         pytest \
-        traitlets==5.9.0 
+        traitlets==5.9.0 \
         qsharp-jupyterlab && \
 # Give permissions to the jovyan user
     chown -R ${USER} ${HOME} && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,9 @@ USER root
 RUN pip install -I --no-cache-dir \
         matplotlib \
         numpy \
-        pytest && \
+        pytest \
+        traitlets==5.9.0 
+        qsharp-jupyterlab && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,7 +16,9 @@ USER root
 RUN pip install -I --no-cache-dir \
         matplotlib \
         numpy \
-        pytest && \
+        pytest \
+        traitlets==5.9.0
```
rohithreddykota:
```suggestion
        traitlets==5.9.0 \
```

Dockerfile
```diff
@@ -85,6 +87,11 @@ RUN cd ${HOME}/ && \
           </configuration>\
     " > ${HOME}/.nuget/NuGet/NuGet.Config
 
+# Install other tutorial-specific packages
+RUN dotnet iqsharp install --user
+RUN cd ${HOME}/QuantumKatas/tutorials/QuantumClassification/ && \
```
rohithreddykota:
```suggestion
RUN cd ${HOME}/tutorials/QuantumClassification/ && \
```

tcNickolas:
This is a known issue related to the deprecation of the Classic QDK and its components, including iqsharp. Classic QDK will be going away completely in June, so we're not investing in improvements to it and its infrastructure, including the Dockerfile for the Classic Katas - the Katas will be used as [https://quantum.microsoft.com/en-us/experience/quantum-katas](https://quantum.microsoft.com/en-us/experience/quantum-katas) going forward.

I will keep this PR open in case somebody wants to run the katas using it in the next two months, but unfortunately, I won't be able to verify the suggested change and merge it into main.

Thank you for your work!

